PR #19, opened 2019-12-17
Implement support for misplaced dot in appended decimal as string

abyss7:


clickhouse/columns/decimal.cpp
```diff
@@ -32,26 +34,49 @@ void ColumnDecimal::Append(const Int128& value) {
...
+            } else {
+                std::advance(end, scale - distance);
+            }
+
+            has_dot = true;
         } else if (*c >= '0' && *c <= '9') {
-            int_value = int_value * 10 + (*c - '0');
+            if (__builtin_mul_overflow(int_value, 10, &int_value) ||
+                __builtin_add_overflow(int_value, *c - '0', &int_value)) {
+                throw std::runtime_error("value is to big for 128-bit integer");
```
alexey-milovidov: to -> too

clickhouse/columns/decimal.cpp
```diff
@@ -32,26 +34,49 @@ void ColumnDecimal::Append(const Int128& value) {
...
         ++c;
     }
 
-    if (c != value.end()) {
-        // TODO: throw exception about symbols after 'minus'
+    if (c != end) {
+        throw std::runtime_error("unexpected symbol '-' in decimal value");
+    }
+
+    while (zeroes) {
```
alexey-milovidov: `zeros` (noun) means "multiple zero digits"
`zeroes` (verb) means "someone zeroes something out" or just a typo.

PR #20, opened 2019-12-19
__int128 usage (c++14 compatibility)

inkrement:
Seems like the used type "__int128" is not full c++14 compatible, but can be used as extension. To avoid warnings during compilation, we explicitly mentioned the usage of this extension.


inkrement:
we have some conflicts - I'll take another look at it.

Issue #30, opened 2020-02-27
Cannot insert data into the table with ' ' (space) in one of its column name.


pakapor:
From one of the example cases, if I put the column name `test id` and run the insert function, the program throws the exception.

Example Code:
```
#include <clickhouse/client.h>
using namespace clickhouse;

/// Initialize client connection.
Client client(ClientOptions().SetHost("localhost"));

/// Create a table.
client.Execute("CREATE TABLE IF NOT EXISTS test.numbers (\`test id\` UInt64, name String) ENGINE = Memory");

/// Insert some values.
try {
	Block block;

	auto id = std::make_shared<ColumnUInt64>();
	id->Append(1);
	id->Append(7);

	auto name = std::make_shared<ColumnString>();
	name->Append("one");
	name->Append("seven");

	block.AppendColumn("test id"  , id);
	block.AppendColumn("name", name);

	client.Insert("test.numbers", block);
} catch (std::exception &e) {
	printf("Error: %s\n", e.what());
}
```

`
Error: DB::Exception: Syntax error: failed at position 33: id,name ) VALUES. Expected one of: token, ClosingRoundBracket, Comma, Dot`

I then tried putting ' " and \` around the column name for the data block, it throws on another line instead.

```block.AppendColumn("`test id`"  , id);```

`Error: DB::Exception: Not found column `test id` in block. There are only columns: test id, name`

PR #31, opened 2020-02-27
Fix Insert into column name with space

pakapor:
Fix issue: #30

Escape special characters and put \`  around the column name.


abyss7:
Thank you for your fix! May you add also an unit-test, please.

Dlougach:
It would be great to have this merged, because the original issue (#30) is painful and has no reasonable workaround.

Enmk:
@pakapor Please merge https://github.com/pakapor/clickhouse-cpp/pull/1 into this. It has some minor simplifications + more checks.

PR #33, opened 2020-03-23
LowCardinality column implementation

Enmk:
Implemented LowCardinality(String) and LowCardinality(FixedString) with native binary serialization and a multitude of tests.

To enable efficient LowCardinality implementation, added some methods to the `Column` interface (and implementation in most derived classes), and added `ItemView` class (see the documentation in sources).

Also, to enable receiving LC columns from the server, updated client TCP protocol revision to 54405 (`DBMS_MIN_REVISION_WITH_LOW_CARDINALITY_TYPE`). That also forced me to update client/server protocol a bit: 
* read server display string and version patch
* send client version patch
* fix parsing of DateTime type with timezone

Known limitations:
* LowCardniality of Nullable is not supported.


ut/utils.h
```diff
@@ -0,0 +1,87 @@
...
+};
+
+template <typename ChronoDurationType>
+class PausableTimer
+{
+public:
+    PausableTimer()
+    {}
+
+    void Start()
```
traceon: Member functions name case is inconsistent with naming of functions in `Timer`. Not significant, just too obvious.

Enmk:
Comparison of populating, inserting and selecting 1'000'000 items performance-wise:

| type | UInt64 | String | FixedString(8) | LowCardinality(String) | LowCardinality(FixedString(8)) |
| ---- | ------ | ------ | -------------- | ---------------------- | ------------------------------ |
| Appending | 10186 | 16580 | 9503 | 136121 | 129960 |
| INSERT | 1465 | 20953 | 1495 | 922 | 964 |
| SELECT | 434942 | 473523 | 448177 | 514830 | 553979 |
| Dictionary size: |  |  |  | 118 | 116 |

Please note: 
* same 8-byte strings were used for both `FixedString(8)` and `LowCardinaility(FixedString(8))`
* same 7-byte strings were used for both `String` and `LowCardinaility(String)`
* dictionary size reflects number of unique items + 1 extra `NULL`-item which is an implementation detail.

clickhouse/types/type_parser.cpp
```diff
@@ -138,6 +210,8 @@ TypeParser::Token TypeParser::NextToken() {
                 return Token{Token::RPar, StringView(cur_++, 1)};
             case ',':
                 return Token{Token::Comma, StringView(cur_++, 1)};
+            case '\'':
+                return Token{Token::Quote, StringView(cur_++, 1)};
```
traceon: I figure, it's better to do quote-to-quote consumption here and generating `Token::String` or some `Token::QuotedString` directly from here. The current approach feels like it somewhat leaks abstractions from lexer (this function) and parser `Parse()`.

clickhouse/columns/array.cpp
```diff
@@ -78,6 +78,13 @@ size_t ColumnArray::Size() const {
     return offsets_->Size();
 }
 
+void ColumnArray::Swap(Column& other) {
+    if (auto col = dynamic_cast<ColumnArray*>(&other)) {
```
traceon: For this and all other `Swap()`'s where the dynamic cast is performed, is it a correct assumption to make, that nothing should happen if the actual type of the argument doesn't match the `*this` type?

traceon: Maybe a dynamic casting of the reference, with a possibility of an exception, is a safer way to do this? (Unless there is a proper way of handling such situations.)

clickhouse/columns/enum.cpp
```diff
@@ -90,6 +90,19 @@ ColumnRef ColumnEnum<T>::Slice(size_t begin, size_t len) {
...
+void ColumnEnum<T>::Swap(Column& other) {
+    if (auto col = dynamic_cast<ColumnEnum<T>*>(&other)) {
+        data_.swap(col->data_);
+    }
+}
+
+template <typename T>
+ItemView ColumnEnum<T>::GetItem(size_t index) const {
+    const T value = data_[index];
+    return ItemView{type_->GetCode(), value};
```
traceon: Just looking at this isolated piece, it feels more safe performance wise to avoid potential creation of unnecessary local copy here. I.e, suggesting changing to this: 
```c++
    return ItemView{type_->GetCode(), data_[index]};
```

clickhouse/columns/nullable.h
```diff
@@ -42,10 +42,41 @@ class ColumnNullable : public Column {
...
+    {
+        return *static_cast<const ColumnNullable &>(col).Nested();
+    }
+
+    return col;
+}
+
+inline Column & removeNullable(Column & col)
+{
+    return const_cast<Column &>(removeNullable(const_cast<const Column&>(col)));
```
traceon: Why `const_cast` if you have an overload for both const and mutable refs? Moreover, maybe better to provide a non-const `Nested()` and don't use `const_cast` at all?

clickhouse/client.cpp
```diff
@@ -563,6 +570,8 @@ void Client::Impl::SendQuery(const std::string& query) {
 
         if (server_info_.revision >= DBMS_MIN_REVISION_WITH_QUOTA_KEY_IN_CLIENT_INFO)
             WireFormat::WriteString(&output_, info.quota_key);
+        if (server_info_.revision >= DBMS_MIN_REVISION_WITH_VERSION_PATCH)
+            WireFormat::WriteUInt64(&output_, info.client_revision);
```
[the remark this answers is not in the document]
Enmk: Server is Ok with that, right now this just plugs the hole, this change is collateral to the PR goal.

If we need the server to know the patch value, we can fix that later

clickhouse/types/typemap.h
```diff
@@ -0,0 +1,33 @@
...
+        return Type::Code::UInt8;
+    } else if constexpr (std::is_signed_v<T> && sizeof(T) == sizeof(std::int16_t)) {
+        return Type::Code::Int16;
+    } else if constexpr (std::is_unsigned_v<T> && sizeof(T) == sizeof(std::uint16_t)) {
+        return Type::Code::UInt16;
+    } else if constexpr (std::is_signed_v<T> && sizeof(T) == sizeof(std::uint32_t)) {
+        return Type::Code::Int32;
+    } else if constexpr (std::is_unsigned_v<T> && sizeof(T) == sizeof(std::int32_t)) {
+        return Type::Code::UInt32;
+    } else if constexpr (std::is_signed_v<T> && sizeof(T) == sizeof(std::int64_t)) {
```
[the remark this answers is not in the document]
Enmk: Ooops, this was not supposed to be committed! Thanks for finding

ut/utils.h
```diff
@@ -0,0 +1,87 @@
...
+    void Restart()
+    {
+        total = ChronoDurationType{0};
+        Start();
+    }
+
+private:
+    Timer<ChronoDurationType> timer;
+    ChronoDurationType total = ChronoDurationType{0};
+    bool paused = false;
```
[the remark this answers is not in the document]
Enmk: why should it be constructed in the started state? IMO both options (started and non-started) are equally frustrating (just for different people).

clickhouse/client.cpp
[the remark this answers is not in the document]
traceon: At least a `client_version_patch` member that is always `0` (even if `client_revision` is also always `0`) would be natural thing to expect here. Like a self-commenting code.

ut/CMakeLists.txt
```diff
@@ -3,11 +3,17 @@ ADD_EXECUTABLE (clickhouse-cpp-ut
...
     client_ut.cpp
     columns_ut.cpp
+    columns_ut.cpp
     socket_ut.cpp
     stream_ut.cpp
-    tcp_server.cpp
+    type_parser_ut.cpp
     type_parser_ut.cpp
     types_ut.cpp
+    types_ut.cpp
```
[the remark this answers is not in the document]
Enmk: at least those are sorted :), fixing...

clickhouse/columns/array.cpp
Enmk: Well, in the perfect world that would throw an exception, but right now it is consistent with the rest of the code. I assume one day we'll implement proper error management.

clickhouse/columns/nullable.h
Enmk: removed that altogether, it is not used anywhere

ut/utils.h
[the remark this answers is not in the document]
traceon: Right, I am not for or against any of those assumptions. I am more concerned with inconsistencies caused by that `bool paused = false;`. It is not constructed started, so it is probably better to be "paused" by default. If I call elapsed on a newly created but not started timer, I'll get some big number. Which is rather confusing. It looks like the running state (`paused == false`) is a property of `Timer` class (could be called `stopped`, and be calculated as (`started_at == 0`)), which can be derived from it by `PausableTimer`.

clickhouse/columns/array.cpp
traceon: I think it's better to change in this PR, because it looks like there is no `dynamic_cast` anywhere else. The change is pretty mechanic though:

Enmk: Yeah, except for `std::dynamic_pointer_cast` [here](https://github.com/ClickHouse/clickhouse-cpp/blob/master/clickhouse/columns/column.h#L23), which is used ALOT.

But I agree, let's change it here, in `Swap()` that is the right thing to do. The rest is coming later.

clickhouse/client.cpp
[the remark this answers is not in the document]
Enmk: Ok, makes sense

ut/utils.h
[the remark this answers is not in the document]
Enmk: Ah, I see, fixed

ut/utils.h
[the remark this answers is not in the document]
Enmk: fixed

ut/CMakeLists.txt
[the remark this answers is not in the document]
Enmk: fixed

ut/client_ut.cpp
```diff
@@ -8,17 +8,16 @@ using namespace clickhouse;
 class ClientCase : public testing::TestWithParam<ClientOptions> {
 protected:
     void SetUp() override {
-        client_ = new Client(GetParam());
+        client_.reset(new Client(GetParam()));
```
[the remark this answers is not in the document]
Enmk: done

clickhouse/types/type_parser.cpp
```diff
@@ -5,8 +5,45 @@
...
+template <typename Container1, typename Container2>
+bool containersMatch(const Container1 & left, const Container2 & right)
+{
+    if (left.size() != right.size())
+        return false;
+
+    const auto l_begin = std::begin(left);
+    const auto l_end = std::end(left);
+    const auto r_begin = std::begin(left);
+    const auto r_end = std::end(left);
```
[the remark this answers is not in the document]
Enmk: removed that piece

clickhouse/types/type_parser.cpp
Enmk: agree, updated implementation of `NextToken()`

clickhouse/columns/array.cpp
Enmk: done

clickhouse/columns/enum.cpp
Enmk: well, that would most likely not make any difference, but for sake of simplicity and brievety, fixed

Enmk:
Updated and fixed all mentioned issues + fixed inconsistent whitespace (at least in code added/updated in this PR)

ut/CMakeLists.txt
```diff
@@ -5,9 +5,13 @@ ADD_EXECUTABLE (clickhouse-cpp-ut
     columns_ut.cpp
     socket_ut.cpp
     stream_ut.cpp
-    tcp_server.cpp
     type_parser_ut.cpp
     types_ut.cpp
+    types_ut.cpp
```
traceon: Duplicate entry.

ut/performance_tests.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    std::cerr << "\n===========================================================" << std::endl;
+    std::cerr << "\t" << ITEMS_COUNT << " items of " << column.Type()->GetName()  << std::endl;
+
+    PausableTimer<std::chrono::microseconds> timer;
+
+    timer.Start();
+    for (size_t i = 0; i < ITEMS_COUNT; ++i)
+    {
+        const auto value = generate(column, i);
```
traceon: Looks like you also measure generation time. That will contaminate the measurements. (in other places too)

ut/performance_tests.cpp
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    EXPECT_EQ(ITEMS_COUNT, column.Size());
+    std::cerr << "Appending:\t" << timer.GetTotalElapsed().count() << " us"
+              << std::endl;
+
+    timer.Reset();
+    timer.Start();
+    EXPECT_NO_FATAL_FAILURE(ValidateColumnItems(column, ITEMS_COUNT));
+    std::cerr << "Accessing (twice):\t" << timer.GetTotalElapsed().count() << " us"
```
traceon: This is also a heavy operation, that will add up to the measurements.

ut/performance_tests.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+        for (int i = 0; i < LOAD_AND_SAVE_REPEAT_TIMES; ++i)
+        {
+            buffer.clear();
+            BufferOutput bufferOutput(&buffer);
+            CodedOutputStream ostr(&bufferOutput);
+
+            timer.Start();
+            column.Save(&ostr);
+            ostr.Flush();
```
traceon: Do you really want to measure flush too?

ut/performance_tests.cpp
```diff
@@ -0,0 +1,233 @@
...
+    std::cerr << "Appending:\t" << timer.GetTotalElapsed().count() << " us"
+        << std::endl;
+
+    EXPECT_NO_FATAL_FAILURE(ValidateColumnItems(column, ITEMS_COUNT));
+
+    {
+        Block block;
+        block.AppendColumn(column_name, column.Slice(0, ITEMS_COUNT));
+
+        timer.Restart();
```
traceon: To minimize the risks of typos, especially in such precise procedure like performance measurement, I think it's better to create new instances of short-lived timer for each sub-case, if you are not going to accumulate measurements. (here and below)

ut/utils.h
```diff
@@ -0,0 +1,87 @@
...
+    void Start()
+    {
+        timer.Restart();
+        paused = false;
+    }
+
+    void Pause()
+    {
+        total += timer.Elapsed();
+        paused = true;
```
traceon: Timer should be stopped, or a check should be added. Consecutive calls to `Pause()` will mess things up.

clickhouse/columns/lowcardinality.cpp
```diff
@@ -0,0 +1,362 @@
+#include "lowcardinality.h"
+
+#include "string.h"
+#include "nullable.h"
+#include "../base/wire_format.h"
```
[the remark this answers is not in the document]
Enmk: Totally agree, but this is out of scope

ut/client_ut.cpp
```diff
@@ -103,6 +102,40 @@ TEST_P(ClientCase, Date) {
...
+    client_->Execute("CREATE TABLE IF NOT EXISTS "
+            "test_clickhouse_cpp.low_cardinality (lc LowCardinality(String)) "
+            "ENGINE = Memory");
+
+    auto lc = std::make_shared<ColumnLowCardinalityT<ColumnString>>();
+
+    const std::vector<std::string> data{{"FooBar", "1", "2", "Foo", "4", "Bar", "Foo", "7", "8", "Foo"}};
+    lc->AppendMany(data);
+
+    block.AppendColumn("lc", lc);
```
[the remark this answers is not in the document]
Enmk: Well, that is somewhat debatable, here we add column to a block, not concatenate columns. But anyway, out of scope.

ut/performance_tests.cpp
Enmk: Yep, but it is very small: no locks, no allocations, calls should be inlined. Also, since those perf tests are ment to be compared in before/after fashion (or against other columns), that tiny “contamination” is cancelled out on comparison.

ut/performance_tests.cpp
Enmk: Yep, since lots of interesting things can happen there, also I am measuring the whole process, the way it is in real cases.

ut/performance_tests.cpp
traceon: If it's tiny then ok.

clickhouse/columns/lowcardinality.cpp
```diff
@@ -0,0 +1,362 @@
...
+    return result;
+}
+
+void ColumnLowCardinality::Swap(Column& other) {
+    auto col = dynamic_cast<ColumnLowCardinality*>(&other);
+    if (!col || !dictionary_column_->Type()->IsEqual(col->dictionary_column_->Type()))
+        return;
+
+    dictionary_column_->Swap(*col->dictionary_column_);
+    index_column_.swap(col->index_column_);
```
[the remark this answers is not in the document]
Enmk: It doesn’t matter here, but for consistency, let’s update it.

ut/performance_tests.cpp
traceon: OK, however potentially LowCardinality + flush, and regular type + flush numbers may differ insignificantly, if flush is heavy. While isolated LowCardinality may be like several times slower than a regular type, and that difference just dwarfed by flush overhead.

ut/client_ut.cpp
```diff
@@ -103,6 +102,40 @@ TEST_P(ClientCase, Date) {
...
+    client_->Insert("test_clickhouse_cpp.low_cardinality", block);
+
+    client_->Select("SELECT lc FROM test_clickhouse_cpp.low_cardinality",
+        [&data](const Block& block)
+        {
+            if (block.GetRowCount() == 0) {
+                return;
```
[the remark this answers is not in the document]
Enmk: Meh, this part is messy. First we receive the “header” block with no rows but with proper columns. Then the data block (possibly more than one), then empty block (no columns nor rows).

This is the way, and it shouldn’t be changed in this PR.
The major Complications are: if we change the API, we mess up user code (which is bad). So at least for now, let’s stick with this less-than perfect API. (Until next major version I guess)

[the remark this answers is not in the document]
traceon: I mean, look at the `TEST_P(ClientCase, Generic)` or `TEST_P(ClientCase, Array)` - they maintain `row` and check it afterwards. Here and in some other test cases this check is not performed, making a `Select()` call, that returns just a single empty block, acceptable.

Enmk:
After the most recent update to implementation and measurements:
| | UInt64 | String | FixedString(8) | LowCardinality(String) | LowCardinality(FixedString(8)) |
| --- | --- | --- | --- | --- | --- |
| Append | 10452 | 17879 | 10984 | 60052 | 61923 |
| INSERT | 1418 | 22340 | 1409 | 732 | 736 |
| SELECT | 4432 | 25419 | 2090 | 459 | 545 |

Tested against localhost server, so the IO lag is diminishable. Tested on 1M entries, time is in microseconds.

ut/utils.h
Enmk: removed that piece

ut/performance_tests.cpp
Enmk: done

ut/performance_tests.cpp
Enmk: It is quite the opposite: serialization/deserialization of `LowCardinality` columns is **MUCH** faster (up to 100 times faster than bare `String` and up to 3 times faster than bare `FixedSring` columns)

ut/performance_tests.cpp
Enmk: Np, fixed

ut/performance_tests.cpp
traceon: Yes, sorry, my bad. But the idea remains the same, observer wont be able so properly see that degree of difference.

ut/client_ut.cpp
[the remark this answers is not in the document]
Enmk: Ok, fixed

clickhouse/columns/lowcardinality.cpp
```diff
@@ -0,0 +1,362 @@
...
+        // Hence in catch-block we assume that dictionary wasn't modified on exception
+        // and there is nothing to rollback.
+
+        appendIndex(iterator->second);
+        if (is_new_item) {
+            AppendToDictionary(dictionary_column_, value);
+        }
+    }
+    catch (...) {
+        removeLastIndex();
```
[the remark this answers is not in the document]
Enmk: Fixed, reverting only if index was added.

traceon:
LGTM

PR #38, opened 2020-04-23
WIP: [Issue-36] Add DateTime64 data type.

tshev:
Going to add tests, and we are good to go.

### DIFF
--- a/clickhouse/columns/date.cpp
+++ b/clickhouse/columns/date.cpp
@@ -113,19 +113,15 @@ ItemView ColumnDateTime::GetItem(size_t index) const {
     return data_->GetItem(index);
 }
 
-
-ColumnDateTime64::ColumnDateTime64()
-    : Column(Type::CreateDateTime64(3ul))
-    , data_(std::make_shared<ColumnDecimal>(18ul, 3ul))
-{}
-
 ColumnDateTime64::ColumnDateTime64(size_t precision)
     : Column(Type::CreateDateTime64(precision))
     , data_(std::make_shared<ColumnDecimal>(18ul, precision))
 {}
 
 
-void ColumnDateTime64::Append(const Int128& value) {
+void ColumnDateTime64::Append(const Int64& value) {
+    // TODO: we need a type, which represents datetime
+    // The precision of Poco.DateTime is not big enough.
     data_->Append(value);
 }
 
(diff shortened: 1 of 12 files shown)


clickhouse/columns/date.cpp
```diff
@@ -113,4 +113,69 @@ ItemView ColumnDateTime::GetItem(size_t index) const {
     return data_->GetItem(index);
 }
 
+
+ColumnDateTime64::ColumnDateTime64()
+    : Column(Type::CreateDateTime64(3ul))
```
tshev, line 118: Maybe we shouldn't provide a default constructor at all or come up with a different default precision. I will be happy to get feedback regarding this line.

clickhouse/columns/date.cpp
```diff
@@ -113,4 +113,69 @@ ItemView ColumnDateTime::GetItem(size_t index) const {
...
+void ColumnDateTime64::Append(const Int128& value) {
+    data_->Append(value);
+}
+
+void ColumnDateTime64::Append(const std::string& value) {
+    data_->Append(value);
+}
+
+
+Int128 ColumnDateTime64::At(size_t n) const {
```
tshev, line 137: I Decided to use Int128 because ColumnDecimal deals with this type. Poco supports https://pocoproject.org/docs/Poco.DateTime.html, but DateTime64 in Clickhouse has a higher resolution than in Poco.

clickhouse/columns/date.cpp
```diff
@@ -113,4 +113,69 @@ ItemView ColumnDateTime::GetItem(size_t index) const {
...
+}
+
+void ColumnDateTime64::Swap(Column& other) {
+    auto& col = dynamic_cast<ColumnDateTime64&>(other);
+    data_.swap(col.data_);
+}
+
+ColumnRef ColumnDateTime64::Slice(size_t begin, size_t len) {
+    auto col = data_->Slice(begin, len)->As<ColumnDecimal>();
+    size_t precision = col->Type()->As<DateTime64Type>()->GetPrecision();
```
Enmk, line 173: that would blow up terribly, do you mean `col->Type()->As<DecimalType>()->GetPrecision()`, also please add one method to `DecimalType`:
```
inline size_t GetPrecision() const { return precision_; }
```

clickhouse/types/types.h
```diff
@@ -144,6 +147,18 @@ class DecimalType : public Type {
     const size_t precision_, scale_;
 };
 
+class DateTime64Type: public Type {
+public:
+    DateTime64Type(size_t precision);
```
Enmk, line 152: please make it `explicit`

asierguti:
What's the status of this PR? I would like to use this feature. Is there any way I could contribute or help anyway?

PR #45, opened 2020-07-20
Implemented reading values of SimpleAggregateFunction

Enmk:
Columns of `SimpleAggregateFunction(func, X)` are now handled properly and can be read from client's code as Column-`X`.

### DIFF
--- a/ut/client_ut.cpp
+++ b/ut/client_ut.cpp
@@ -284,6 +284,34 @@ TEST_P(ClientCase, Numbers) {
     EXPECT_EQ(100000U, num);
 }
 
+TEST_P(ClientCase, SimpleAggregateFunction) {
+    client_->Execute("DROP TABLE IF EXISTS test_clickhouse_cpp.SimpleAggregateFunction");
+    client_->Execute(
+            "CREATE TABLE IF NOT EXISTS test_clickhouse_cpp.SimpleAggregateFunction (saf SimpleAggregateFunction(sum, UInt64))"
+            "ENGINE = Memory");
+
+    constexpr size_t EXPECTED_ROWS = 10;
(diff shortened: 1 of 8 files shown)


ut/columns_ut.cpp
```diff
@@ -258,3 +259,11 @@ TEST(ColumnsCase, LowCardinalityString_Save) {
         EXPECT_EQ(col.At(i), foobar(i)) << " at pos: " << i;
     }
 }
+
+TEST(ColumnsCase, CreateSimpleAggregateFunction) {
+    auto col = CreateColumnByType("SimpleAggregateFunction(funt, Int32");
```
traceon, line 264: `SimpleAggregateFunction(funt, Int32` (without closing parentheses)? Does this test passes? If yes, then why? Should it pass (i.e. should type parser tolerate this)?

ut/client_ut.cpp
```diff
@@ -284,6 +297,40 @@ TEST_P(ClientCase, Numbers) {
...
+        return;
+    }
+
+    client_->Execute("DROP TABLE IF EXISTS test_clickhouse_cpp.SimpleAggregateFunction");
+    client_->Execute(
+            "CREATE TABLE IF NOT EXISTS test_clickhouse_cpp.SimpleAggregateFunction (saf SimpleAggregateFunction(sum, UInt64))"
+            "ENGINE = Memory");
+
+    constexpr size_t EXPECTED_ROWS = 10;
+    client_->Execute("INSERT INTO test_clickhouse_cpp.SimpleAggregateFunction (saf) SELECT number FROM system.numbers LIMIT 10");
```
traceon, line 313: I would say `... LIMIT " + std::to_string(EXPECTED_ROWS);`

PR #47, opened 2020-08-11
Cherry-pick https://github.com/artpaul/clickhouse-cpp/pull/112 

filimonov:
try to reconnect send_retries times before throw in RetryGuard (artpul#112)
See  https://github.com/artpaul/clickhouse-cpp/pull/112

### DIFF
--- a/clickhouse/client.h
+++ b/clickhouse/client.h
@@ -40,7 +40,7 @@ struct ClientOptions {
     /// Hostname of the server.
     DECLARE_FIELD(host, std::string, SetHost, std::string());
     /// Service port.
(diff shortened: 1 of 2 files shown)


clickhouse/client.h
```diff
@@ -40,7 +40,7 @@ struct ClientOptions {
     /// Hostname of the server.
     DECLARE_FIELD(host, std::string, SetHost, std::string());
     /// Service port.
-    DECLARE_FIELD(port, int, SetPort, 9000);
+    DECLARE_FIELD(port, unsigned int, SetPort, 9000);
```
traceon, line 43: Won't this cause parsing issues anywhere?

PR #50, opened 2020-09-03
Add DateTime64 data type.

Enmk:
Based on #38 
New Column type: `ColumnDateTime64`
Sligtly updated some underlying code
Added unit and integrration tests for ColumnDateTime64

Closes #38

### DIFF
--- a/clickhouse/columns/decimal.cpp
+++ b/clickhouse/columns/decimal.cpp
@@ -91,7 +91,6 @@ Int128 ColumnDecimal::At(size_t i) const {
             return data_->As<ColumnInt128>()->At(i);
         default:
             throw std::runtime_error("Invalid data_ column type in ColumnDecimal");
-            return 0;
     }
 }
 
--- a/clickhouse/columns/decimal.cpp
+++ b/clickhouse/columns/decimal.cpp
@@ -1,6 +1,6 @@
 #include "decimal.h"
 
-#include <iostream>
+#include <cassert>
 
 namespace clickhouse {
 
@@ -83,12 +83,16 @@ void ColumnDecimal::Append(const std::string& value) {
 }
 
 Int128 ColumnDecimal::At(size_t i) const {
-    if (data_->Type()->GetCode() == Type::Int32) {
-        return static_cast<Int128>(data_->As<ColumnInt32>()->At(i));
-    } else if (data_->Type()->GetCode() == Type::Int64) {
-        return static_cast<Int128>(data_->As<ColumnInt64>()->At(i));
-    } else {
-        return data_->As<ColumnInt128>()->At(i);
+    switch (data_->Type()->GetCode()) {
+        case Type::Int32:
+            return static_cast<Int128>(data_->As<ColumnInt32>()->At(i));
+        case Type::Int64:
+            return static_cast<Int128>(data_->As<ColumnInt64>()->At(i));
+        case Type::Int128:
+            return data_->As<ColumnInt128>()->At(i);
+        default:
+            assert(false && "Invalid data_ column type in ColumnDecimal");
+            return 0;
     }
 }
 
@@ -127,6 +131,16 @@ void ColumnDecimal::Swap(Column& other) {
 
 ItemView ColumnDecimal::GetItem(size_t index) const {
     return data_->GetItem(index);
+}
+
+size_t ColumnDecimal::GetScale() const
+{
+    return type_->As<DecimalType>()->GetScale();
+}
+
+size_t ColumnDecimal::GetPrecision() const
+{
+    return type_->As<DecimalType>()->GetPrecision();
 }
 
 }
--- a/clickhouse/columns/decimal.cpp
+++ b/clickhouse/columns/decimal.cpp
@@ -1,7 +1,5 @@
 #include "decimal.h"
 
-#include <cassert>
-
 namespace clickhouse {
 
 ColumnDecimal::ColumnDecimal(size_t precision, size_t scale)
@@ -16,8 +14,9 @@ ColumnDecimal::ColumnDecimal(size_t precision, size_t scale)
     }
 }
 
-ColumnDecimal::ColumnDecimal(TypeRef type)
-    : Column(type)
+ColumnDecimal::ColumnDecimal(TypeRef type, ColumnRef data)
+    : Column(type),
+      data_(data)
 {
 }
 
@@ -91,7 +90,7 @@ Int128 ColumnDecimal::At(size_t i) const {
         case Type::Int128:
             return data_->As<ColumnInt128>()->At(i);
         default:
-            assert(false && "Invalid data_ column type in ColumnDecimal");
+            throw std::runtime_error("Invalid data_ column type in ColumnDecimal");
             return 0;
     }
 }
@@ -119,9 +118,8 @@ size_t ColumnDecimal::Size() const {
 }
 
 ColumnRef ColumnDecimal::Slice(size_t begin, size_t len) {
-    std::shared_ptr<ColumnDecimal> slice(new ColumnDecimal(type_));
-    slice->data_ = data_->Slice(begin, len);
-    return slice;
+    // coundn't use std::make_shared since this c-tor is private
+    return ColumnRef{new ColumnDecimal(type_, data_->Slice(begin, len))};
 }
 
 void ColumnDecimal::Swap(Column& other) {
--- a/clickhouse/columns/numeric.cpp
+++ b/clickhouse/columns/numeric.cpp
@@ -10,9 +10,9 @@ ColumnVector<T>::ColumnVector()
 }
 
 template <typename T>
-ColumnVector<T>::ColumnVector(const std::vector<T>& data)
+ColumnVector<T>::ColumnVector(std::vector<T> data)
     : Column(Type::CreateSimple<T>())
-    , data_(data)
+    , data_(std::move(data))
 {
 }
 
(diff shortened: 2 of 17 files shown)


clickhouse/columns/decimal.cpp
```diff
@@ -83,12 +83,16 @@ void ColumnDecimal::Append(const std::string& value) {
...
-        return data_->As<ColumnInt128>()->At(i);
+    switch (data_->Type()->GetCode()) {
+        case Type::Int32:
+            return static_cast<Int128>(data_->As<ColumnInt32>()->At(i));
+        case Type::Int64:
+            return static_cast<Int128>(data_->As<ColumnInt64>()->At(i));
+        case Type::Int128:
+            return data_->As<ColumnInt128>()->At(i);
+        default:
+            assert(false && "Invalid data_ column type in ColumnDecimal");
```
traceon, line 94: Please, do not use `assert()` for failures that may happen during normal operation (or depend on input data) and not only in testing (which means pretty much never.) These asserts will be no-op in release builds, leaving the code without that protection.

clickhouse/columns/numeric.cpp
```diff
@@ -10,9 +10,9 @@ ColumnVector<T>::ColumnVector()
 }
 
 template <typename T>
-ColumnVector<T>::ColumnVector(const std::vector<T>& data)
+ColumnVector<T>::ColumnVector(std::vector<T> data)
     : Column(Type::CreateSimple<T>())
-    , data_(data)
+    , data_(std::move(data))
```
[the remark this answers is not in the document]
traceon: If there are cases, when the passed argument is a temporary that could be moved-in, then better to add a move c-tor: `ColumnVector(std::vector<T>&& data)`

ut/client_ut.cpp
```diff
@@ -631,6 +658,71 @@ TEST_P(ClientCase, Decimal) {
...
+    client_->Select("SELECT dt FROM test_clickhouse_cpp.datetime64",
+        [&total_rows, &data](const Block& block) {
+            total_rows += block.GetRowCount();
+            if (block.GetRowCount() == 0) {
+                return;
+            }
+
+            ASSERT_EQ(1U, block.GetColumnCount());
+            if (auto col = block[0]->As<ColumnDateTime64>()) {
+                ASSERT_EQ(data.size(), col->Size());
```
traceon, line 715: Does this assume, that you always read an entire `data`-sized block here?

ut/columns_ut.cpp
```diff
@@ -114,6 +152,135 @@ TEST(ColumnsCase, DateAppend) {
...
+    ASSERT_EQ(0u, column->Size());
+    ASSERT_EQ(data.size(), column2->Size());
+
+    // Clearing doesn't affects precision value.
+    ASSERT_EQ(6ul, column->GetPrecision());
+
+    for (size_t i = 0; i < data.size(); ++i) {
+        ASSERT_EQ(data[i], column2->At(i));
+    }
+}
```
traceon, line 224: This checks only that one direction is OK after the swap, right? One typical bug of a `swap` implementation, is when after a swap the destinations contains what was in the source, but the source is left empty.

ut/columns_ut.cpp
```diff
@@ -114,6 +152,135 @@ TEST(ColumnsCase, DateAppend) {
...
+    }
+
+    ASSERT_EQ(data.size(), column->Size());
+    for (size_t i = 0; i < data.size(); ++i) {
+        ASSERT_EQ(data[i], column->At(i));
+    }
+}
+
+TEST(ColumnsCase, DateTime64_Clear) {
+    auto column = std::make_shared<ColumnDateTime64>(6ul);
```
traceon, line 188: It is worth adding `ASSERT_EQ(0u, column->Size());` to ensure that a column is created empty.

ut/columns_ut.cpp
```diff
@@ -114,6 +152,135 @@ TEST(ColumnsCase, DateAppend) {
...
+        auto slice = column->Slice(offset, count)->As<ColumnDateTime64>();
+
+        ASSERT_EQ(count, slice->Size());
+        ASSERT_EQ(column->GetPrecision(), slice->GetPrecision());
+
+        for (size_t i = offset; i < offset + count; ++i) {
+            ASSERT_EQ(data[i], slice->At(i - offset));
+        }
+    }
+}
```
traceon, line 275: What happens on out-of-bound slices (5 cases)?

traceon (changes requested):
Left comments.

clickhouse/types/type_parser.cpp
```diff
@@ -9,6 +9,7 @@
 namespace clickhouse {
 
 bool TypeAst::operator==(const TypeAst & other) const {
+    // TODO: operator == must be a friend function.
```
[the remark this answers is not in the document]
Enmk: That is a leftovers from previous PR from another author, and I guess we will never figure out why... Let me remove it

ut/client_ut.cpp
```diff
@@ -712,7 +712,6 @@ TEST_P(ClientCase, DateTime64) {
 
             ASSERT_EQ(1U, block.GetColumnCount());
             if (auto col = block[0]->As<ColumnDateTime64>()) {
-                ASSERT_EQ(data.size(), col->Size());
                 for (size_t i = 0; i < col->Size(); ++i) {
                     EXPECT_EQ(data[i], col->At(i)) << " at index: " << i;
```
traceon, line 716: Are the indices absolute? Or they are offsets within the current block?
If the latter, then the code still relies on the assumption that the entire data will be received in the [first] block. Maybe, technically it is what will always happen here, but if the data is set to size of, e.g., `100 000 000`, will this code still be correct?

clickhouse/columns/decimal.cpp
```diff
@@ -91,7 +90,7 @@ Int128 ColumnDecimal::At(size_t i) const {
         case Type::Int128:
             return data_->As<ColumnInt128>()->At(i);
         default:
-            assert(false && "Invalid data_ column type in ColumnDecimal");
+            throw std::runtime_error("Invalid data_ column type in ColumnDecimal");
             return 0;
```
traceon, line 94: Extra return, looks like.

ut/columns_ut.cpp
```diff
@@ -274,6 +274,23 @@ TEST(ColumnsCase, DateTime64_Slice) {
...
+    // Non-Empty slice on empty column
+    EXPECT_EQ(0u, column->Slice(0, 10)->Size());
+
+    const auto data = MakeDateTime64s();
+    for (const auto & v : data) {
+        column->Append(v);
+    }
+
+    EXPECT_EQ(column->Slice(0, data.size() + 1)->Size(), data.size());
+    EXPECT_EQ(column->Slice(data.size() + 1, 1)->Size(), 0u);
```
traceon, line 291: Maybe, also this?

```c++
     EXPECT_EQ(column->Slice(data.size() / 2, data.size() / 2 + 2)->Size(), data.size()); 
```

traceon (commented):
Some comments.

ut/columns_ut.cpp
Enmk: fixed

ut/columns_ut.cpp
Enmk: We check that in other tests.

ut/columns_ut.cpp
Enmk: Added tests

ut/client_ut.cpp
Enmk: Fixed

ut/client_ut.cpp
Enmk: good note, fixed

clickhouse/columns/decimal.cpp
Enmk: Ok

clickhouse/columns/numeric.cpp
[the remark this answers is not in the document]
Enmk: Ok, no problem, I guess it is time for me to re-read core guidelines.

clickhouse/columns/numeric.cpp
```diff
@@ -10,7 +10,14 @@ ColumnVector<T>::ColumnVector()
...
+ColumnVector<T>::ColumnVector(const std::vector<T> & data)
+    : Column(Type::CreateSimple<T>())
+    , data_(data)
+{
+}
+
+template <typename T>
+ColumnVector<T>::ColumnVector(std::vector<T> && data)
     : Column(Type::CreateSimple<T>())
     , data_(data)
```
traceon, line 22: Then, this one should be passed further as an rvalue here: `, data_(std::move(data))`

ut/client_ut.cpp
traceon: I mean, it it correct to use `i` that iterates over `col` ans an index for `data`?

traceon (commented):
One more.

ut/performance_tests.cpp
```diff
@@ -80,7 +80,16 @@ class ColumnPerformanceTest : public ::testing::Test {
...
 
+// Turns out this is the easiest way to skip test with current version of gtest
+#ifndef NDEBUG
+#  define SKIP_IN_DEBUG_BUILDS() do { std::cerr << "Test skipped...\n"; return; } while(0)
+#else
+#  define SKIP_IN_DEBUG_BUILDS() (void)(0)
+#endif
+
 TYPED_TEST_P(ColumnPerformanceTest, SaveAndLoad) {
+    SKIP_IN_DEBUG_BUILDS();
```
traceon, line 91: FYI, Google Test has `GTEST_SKIP()` starting from some version, and, alternatively, you can prepend `DISABLED_` to the name of a test to disable it (should be available in the bundled `gtest` repo). Not an issue here, though.

traceon (approved):

PR #53, opened 2020-09-24
Fix of LowCardinality bug: empty string converted to other value on INSERT

Enmk:
Fix of empty string becoming a first non-empty string of ColumnLowCardinality;
Added some tests to make sure that doesn't happen again.


clickhouse/columns/lowcardinality.cpp
```diff
@@ -365,7 +365,7 @@ void ColumnLowCardinality::AppendNullItemToEmptyColumn()
 
     const auto null_item = GetNullItemForDictionary(dictionary_column_);
     AppendToDictionary(*dictionary_column_, null_item);
-    unique_items_map_.emplace(computeHashKey(null_item), dictionary_column_->Size());
+    unique_items_map_.emplace(computeHashKey(null_item), 0);
```
Enmk, line 368: Here it is, where the bug was introduced (by me)

traceon (approved):
LGTM

PR #54, opened 2020-09-30
Add TCP_NODELAY option

oraoto:
During benchmark, I found there are significant delay when the block is smaller than the buffer size (defaults to 8192 bytes).

Test program:

``` cpp
#include <clickhouse/client.h>
#include <cstdlib>
#include <chrono>
#include <iostream>

using namespace clickhouse;

int main(int argc, char **argv)
{
    Client client(ClientOptions().SetHost("localhost").SetPassword("qwerty"));

    client.Execute("CREATE TABLE IF NOT EXISTS test.delay (id UInt8) ENGINE = Memory");

    int count = atoi(argv[1]); // count

    for (int i = 0; i < 10; i++) {
	Block block;

	auto id = std::make_shared<ColumnUInt8>();

	for (int j = 0; j < count; j++) {
	    id->Append(65);
	}
	block.AppendColumn("id"  , id);

	auto start = std::chrono::steady_clock::now();
	client.Insert("test.delay", block);
	auto end = std::chrono::steady_clock::now();

	std::cout << std::chrono::duration_cast<std::chrono::microseconds>(end - start).count() << " ms" << std::endl;
    }

}
```

When data (8171 bytes + some header) is larger than bufffer:

``` shell
./test-insert-delay 8171
434 ms
567 ms
366 ms
252 ms
354 ms
265 ms
630 ms
237 ms
234 ms
181 ms
````

When data is  smaller than bufffer, there are significant delay:

``` shell
./test-insert-delay 8170
213 ms
155 ms
43767 ms
43270 ms
46539 ms
46627 ms
43114 ms
43273 ms
394 ms
42893 ms
```

To resolve this problem, I add an option to enable TCP_NODELAY. It's enabled by default, same as [clickhouse-client](https://github.com/ClickHouse/ClickHouse/blob/c0d1416bbd564fa1e9b3ae5e58b1a5785b5d891a/src/Client/Connection.cpp#L91).

### DIFF
--- a/clickhouse/base/socket.cpp
+++ b/clickhouse/base/socket.cpp
@@ -169,6 +169,11 @@ void SocketHolder::SetTcpKeepAlive(int idle, int intvl, int cnt) noexcept {
 #endif
 }
 
+void SocketHolder::SetTcpNoDelay(bool nodelay) noexcept {
+    int val = nodelay;
+    setsockopt(handle_, IPPROTO_TCP, TCP_NODELAY, &val, sizeof(val));
+}
+
 SocketHolder& SocketHolder::operator = (SocketHolder&& other) noexcept {
     if (this != &other) {
         Close();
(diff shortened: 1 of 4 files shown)


oraoto:
Well, the time scale should be `μs`, which means Nagle's algorithm adds about 40 ms delay.

PR #71, opened 2020-11-18
Fix click house install in travis ci script

Enmk:
Updated ClickHouse install script to match one from: https://clickhouse.tech/#quick-start


filimonov (approved):
LGTM

PR #66, opened 2020-11-10
use absl for int128

filimonov:
Backport 
https://github.com/artpaul/clickhouse-cpp/pull/119

See
https://github.com/artpaul/clickhouse-cpp/issues/113
https://github.com/artpaul/clickhouse-cpp/issues/116

PR #72, opened 2020-11-25
Force gnu++17 for building (for is_fundamental_v<__int128>==1)

azat:
Since:
- `-std=gnu++17: std::is_fundamental_v<__int128> == 1`
- `-std=c++17:   std::is_fundamental_v<__int128> == 0`

And there is code that requires __int128 "to be fundamental"


azat:
Although it seems that https://github.com/ClickHouse/clickhouse-cpp/pull/66 supersedes this patch

Enmk:
fixed in #66 and #81

PR #73, opened 2020-12-15
Fix for PR #66

nshishov:
Fix for PR #66 
Explicit conversion required for 128 to 64 bit conversation


Enmk:
fixed in #66 and #81

PR #78, opened 2021-02-28
Added declaration of Int128 (fixes segfault)

kolya7k:
* Added declaration of Int128
Without it was a segfault, now a proper std::runtime_error "unsupported column type: Int128" thrown

* Added CLion ignores
* Some style & unused fixes from Clang-Tidy
* Implemented InsertQuery and InsertData based on https://github.com/aiwhj/clickhouse-cpp/commit/bab28bcb5a509d80b8e2e0c7e89512446283dde5
* Made "Query query" "const Query& query" to prevent copying each call


Enmk:
Hi! Thank you for the PR, but could you please split it up? Right now it is a lump of unrelated features.
Also, it begs for tests (at least for segfault part).

PR #87, opened 2021-04-16
Allow clients that lack absl to built and use the library

### LINKED ISSUE #86: relative import error
I did 
```
$ cmake ..
$ make
$ sudo make install
```

```In file included from /usr/local/include/clickhouse/columns/column.h:5,
```

Enmk:
Moved Int128 definition to a separate header
Added few extra methods to ColumnDecimal that do not rely on Int128
Ensured that client programs can be built without absl headers present.

closes: #86

### DIFF
--- a/clickhouse/types/types.h
+++ b/clickhouse/types/types.h
@@ -1,12 +1,15 @@
 #pragma once
 
-#include "absl/numeric/int128.h"
-
 #include <map>
 #include <memory>
 #include <string>
 #include <vector>
 #include <stdexcept>
+
+namespace absl
+{
+class int128;
+}
 
 namespace clickhouse {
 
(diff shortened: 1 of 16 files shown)


clickhouse/types/types.h
```diff
@@ -1,13 +1,16 @@
...
 #include <memory>
 #include <string>
 #include <vector>
 #include <stdexcept>
 
+namespace absl
+{
+class int128;
+}
+
```
traceon, line 13: Isn't it better to move all `Int128` specific code and support into `clickhouse/types/int128.h` so that when user doesn't include that header he can't even refer to any type related/implemented via `Int128`?

traceon (commented):
Some comments inline.

slabko:

Issue #88, opened 2021-05-05
ipv4 column append problem


BozaB:
Dear clickhouse-cpp team,

The following function in  _clickhouse-cpp/clickhouse/columns/ip4.cpp_  is always thrown an exception, when ipv4 address is 255.255.255.255:
```c++
void ColumnIPv4::Append(const std::string& str) {
    in_addr_t addr = inet_addr(str.c_str());
    if (addr == INADDR_NONE) {
        throw std::runtime_error("invalid IPv4 format, ip: " + str);
    }
    data_->Append(htonl(addr));
}
```
Because inet_addr function is obsolete, and this is a known issue. https://www.gnu.org/software/libc/manual/html_node/Host-Address-Functions.html

The possible solution is the following:
```c++
void ColumnIPv4::Append(const std::string& str) {
    struct in_addr addr;
    if (inet_aton(str.c_str(),&addr) == 0) {
        throw std::runtime_error("invalid IPv4 format, ip: " + str);
    }
    data_->Append(htonl(addr.s_addr));
}
```
Please solve this problem.

Thank you and best regards,
BB

filimonov:
Can you send a PR with that change? (this way you will be an author of that :) )

Enmk:
Closed by #148

PR #101, opened 2021-08-26
fix:ipv4 column append problem

ghost:
#88 fix:ipv4 column append problem.author:  BozaB.

### DIFF
--- a/tests/simple/main.cpp
+++ b/tests/simple/main.cpp
@@ -442,6 +442,7 @@ inline void IPExample(Client &client) {
 
         auto v4 = std::make_shared<ColumnIPv4>();
         v4->Append("127.0.0.1");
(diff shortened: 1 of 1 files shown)


BozaB:
Hellozhuzhu, thank you! :)

filimonov:
Great, can you also add the problematic ip to the test?

https://github.com/ClickHouse/clickhouse-cpp/blob/1415b5936a2ac2f084850b09057e05fb5798b2f1/tests/simple/main.cpp#L444

hellozhuzhu:
Of course.

Enmk (changes requested):
Please sign CLA and write some unit tests, `simple/main.cpp` is not built/run in CI/CD

PR #104, opened 2021-09-14
Implemented OnFinish callback in Query and fixed some clang-tidy issues

MikhailNazarov:
When implementing a rest-api, it is necessary to send a response after receiving the full data from the clickhouse, the request completion event is sorely lacking. The OnFinish event does not invokes any callbacks. I implemented this event by analogy with others. Also corrected the clang-tidy issues for Query

### DIFF
--- a/clickhouse/query.h
+++ b/clickhouse/query.h
@@ -6,6 +6,7 @@
 #include <functional>
 #include <memory>
 #include <string>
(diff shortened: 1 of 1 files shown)


MikhailNazarov:
It’s not actual due to single-threaded work of the client

PR #110, opened 2021-10-30
Reduced memory overhead of preparing LZ4-compressed data for server.

Enmk:
Do not compress a whole serialized block, but instead only a reasonable-sized chunk.
This removes some temporary buffers and reduces memory pressure.

Also minor refactoring:
- moved all serialization-format code to WireFormat class.
- removed CodedOutputStream and CodedInputStream classes.

# Memory usage
Test executed against a single block of 5 columns by `INSERT`ing and `SELECT`ing rows back from the server. Both binaries were built in `RelWithDebInfo` mode.

* `initial` - almost right after the program start
* `before INSERTing` - a moment when the block is prepared in memory, but before actual call to `Client::Insert`
* `after INSERTing` - right after `Client::Insert`, NOTE: original Block is still in memory for validation.
* `after SELECTing` - right after `client.Select("SELECT * FROM ...")`, NOTE: original Block is still in memory for validation.

## Original implementation
```
// 10M rows:
Block of: U64 UInt64, S String, FS FixedString(8), LCS LowCardinality(String), LCFS LowCardinality(FixedString(8)) with 10000000rows
RSS
                     initial    value :    4550656      peak:    4550656
            before INSERTing    value :  476459008      peak:  503341056
             after INSERTing    value :  476602368      peak: 1117761536
             after SELECTing    value :  545083392      peak: 1277018112

// 100M rows:
Block of: U64 UInt64, S String, FS FixedString(8), LCS LowCardinality(String), LCFS LowCardinality(FixedString(8)) with 100000000rows
unknown file: Failure
C++ exception with description "DB::Exception: Unexpected packet Data received from client" thrown in the test body.
```

## Version in this PR
```
// 10M rows:
Block of: U64 UInt64, S String, FS FixedString(8), LCS LowCardinality(String), LCFS LowCardinality(FixedString(8)) with 10000000rows
RSS
                     initial    value :    5767168      peak:    5767168
            before INSERTing    value :  477839360      peak:  504721408
             after INSERTing    value :  477839360      peak:  504721408
             after SELECTing    value :  546471936      peak: 1278402560

// 100M Rows
Block of: U64 UInt64, S String, FS FixedString(8), LCS LowCardinality(String), LCFS LowCardinality(FixedString(8)) with 100000000rows
RSS
                     initial    value :    5857280      peak:    5857280
            before INSERTing    value : 4714172416      peak: 4850671616
             after INSERTing    value : 4714172416      peak: 4850671616
             after SELECTing    value : 5696409600      peak: 12135694336
```

## Comparison
Since the original version failed to insert 100M rows, we are going to compare 10M rows memory usage.
As you can see, original implementation peaks to **1117761536** bytes upon insertion, **503341056** of whose are related to the original `Block` residing in memory, that is **1117761536 - 503341056 = 614420480 bytes (~0.57Gib)** of memory used only to INSERT and send data to the server.

The modified implementation uses an insignificant amount of extra memory, untraceable with the current approach. Moreover, it is undetectable even upon insertion of 100M rows.

## Conclusion
Modified version (presented in this PR) uses `O(1)` extra memory, vs `O(n)` for the original version.


traceon:
@Enmk I see some seemingly conflicting changes between this and #109. Let's merge #109 before fully reviewing this one.

clickhouse/base/wire_format.cpp
```diff
@@ -0,0 +1,105 @@
+#include "wire_format.h"
+
+#include "input.h"
+#include "output.h"
+
+#include <stdexcept>
+
+namespace {
+static const int MAX_VARINT_BYTES = 10;
+}
```
traceon, line 10:
```suggestion
namespace {
constexpr int MAX_VARINT_BYTES = 10;
}
```
No need in static with unnamed namespaces.

clickhouse/base/wire_format.cpp
```diff
@@ -0,0 +1,105 @@
...
+        p += read_previously;
+        len -= read_previously;
+    }
+
+    // true if all was read successfully
+    return !len;
+//    if (len) {
+//        throw std::runtime_error("Failed to read " + std::to_string(original_len)
+//                + " bytes, only read " + std::to_string(original_len - len));
+//    }
```
traceon, line 30: Let's either incorporate this check or remove.

clickhouse/base/wire_format.cpp
```diff
@@ -0,0 +1,105 @@
...
+        throw std::runtime_error("Failed to write " + std::to_string(original_len)
+                + " bytes, only written " + std::to_string(original_len - len));
+    }
+}
+
+bool WireFormat::ReadVarint64(InputStream* input, uint64_t* value) {
+    *value = 0;
+
+    for (size_t i = 0; i < MAX_VARINT_BYTES; ++i) {
+        uint8_t byte;
```
traceon, line 55: Uninitialized.

clickhouse/base/wire_format.cpp
```diff
@@ -0,0 +1,105 @@
...
+        if (!value) {
+            break;
+        }
+    }
+
+    WriteAll(output, bytes, size);
+}
+
+bool WireFormat::SkipString(InputStream* input) {
+    uint64_t len;
```
traceon, line 93: Uninitialized.

traceon (changes requested):
Left inline comments.

traceon:
Could you please provide some numbers of improvements? Like "before vs after" perf tests results, etc.

clickhouse/base/input.cpp
```diff
@@ -5,6 +5,21 @@
 
 namespace clickhouse {
 
+bool ZeroCopyInput::Skip(size_t bytes) {
+    while (bytes > 0) {
+        const void* ptr;
```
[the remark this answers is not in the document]
Enmk: yep, it is overwritten in the next step, and not used at all afterward.

clickhouse/base/wire_format.cpp
Enmk: yep, it is overwritten right afterward and there is no reason to initialize it.

clickhouse/base/wire_format.cpp
Enmk: yep, it is overwritten right afterward and there is no reason to initialize it.

clickhouse/base/coded.cpp
```diff
@@ -4,97 +4,27 @@
 
 namespace clickhouse {
 
-static const int MAX_VARINT_BYTES = 10;
+//static const int MAX_VARINT_BYTES = 10;
```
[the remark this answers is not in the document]
Enmk: yep, definitely

clickhouse/base/compressed.cpp
```diff
@@ -95,4 +102,73 @@ bool CompressedInput::Decompress() {
...
+
+    destination_->Flush();
+}
+
+void CompressedOutput::PreallocateCompressBuffer(size_t input_size) {
+    const auto estimated_compressed_buffer_size = LZ4_compressBound(input_size);
+    if (estimated_compressed_buffer_size <= 0)
+        throw std::runtime_error("Failed to estimate compressed buffer size, LZ4 error: " + std::to_string(estimated_compressed_buffer_size));
+
+    compressed_buffer_.resize(estimated_compressed_buffer_size + HEADER_SIZE + EXTRA_COMPRESS_BUFFER_SIZE);
```
traceon, line 171: This is a low hanging fruit for optimization: resize without initialization.

PR #116, opened 2021-11-23
Allow macOS builds to compile and run SSL tests

traceon:
Assuming OpenSSL from Homebrew is installed.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -13,9 +13,10 @@ on:
       extra_install:
         required: false
         type: string
-      gtest_args:
+      gtest_filter:
         required: false
         type: string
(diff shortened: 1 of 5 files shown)


Enmk (dismissed):
Please add a workflow that runs the TLS-related tests on macOS.

Enmk (changes requested):
Please add a workflow that runs the TLS-related tests on macOS.

traceon:
I've added a similar workflow for macOS + TLS but, I guess, `Client/*` of tests won't run, since there is no server available for macOS workflows.

Also, probably need to merge this even with failed workflows, and then fix with a separate PR, due to workflow reuse, which probably reference the `master` branch content.

PR #122, opened 2021-12-03
make client threadsafe

### LINKED ISSUE #37: Thread safety of Client
It seems Client class is not thread safe - i.e. sending concurrent queries from different threads might break. Can you please either make it thread-safe or at least document the current state of things.

caiowakamatsuahrefs:
closes #37

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -15,6 +15,7 @@
 #include <vector>
 #include <sstream>
 #include <stdexcept>
+#include <mutex>
 
 #if defined(WITH_OPENSSL)
 #include "base/sslsocket.h"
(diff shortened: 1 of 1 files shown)


caiowakamatsuahrefs:
@Enmk @Vertver @traceon ping

Enmk (changes requested):
IMO it is best to create a thread-safe wrapper around `Client` rather than making it a thread-safe by-deafult. That would allow users to choose between thread-safe and non-thread-safe-but-more-performant APIs.

filimonov:
+1. Saying that it's threadsafe when actually only a single thread can work with it (and others simply wait for locks) - sounds unfair. I would just add to documentation that client is single threaded and not thread-safe.

PR #131, opened 2022-01-06
Add support for MSVC

### LINKED ISSUE #127: 'Failed to connect' exception using MSVC
Hello,

I was using MSVC to build clickhousecpp (version 2.0.0) on Windows, and the static library was generated successfully. However, an exception of ```failed to connect``` occurred when I run the **simple-test** project. There is a clickhouse server running on my localhost, exposing the 9000 port. I test the connection with python ```clickhouse_driver``` and everything is ok.

By the way, I encountered a problem with ```signbit: ambiguous call to overloaded function``` when I was building the clickhousecpp library. The problem was solved according to [#100](https://github.com/ClickHouse/clickhouse-cpp/issues/100), and I don't know whether it is an official solution.

Is there any idea of this issue?

Thanks

### LINKED ISSUE #130: Add github actions for windows / MSVC build

filimonov:
* github action for msvc added. Closes #130
* there is an issue with dynamic linking & DLL paths when compiling tests. I've switched tests to static linking
* added tls offoader proxy for windows tests, mimicking local clickhouse by tunneling the queries to a cloud instance of clickhouse
* get proper status codes via WSAGetLastError for windows, fixes #127 

Known issues:
- no support for secure connections (need to reimplement secure_socket with sspi.h and other windows thing). Maybe can work with openssl (not tested).
- there are some warnings at compile time

### DIFF
--- a/.github/workflows/windows_msvc.yml
+++ b/.github/workflows/windows_msvc.yml
@@ -1,0 +1,60 @@
+name: Windows
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  BUILD_TYPE: Release
+  GTEST_FILTER: --gtest_filter=-"*"
+  CLICKHOUSE_USER: clickhouse_cpp_cicd
+  CLICKHOUSE_PASSWORD: clickhouse_cpp_cicd
+  #
+  # CLICKHOUSE_HOST: localhost
+  # CLICKHOUSE_PORT: 9000
+  # CLICKHOUSE_USER: default
+  # CLICKHOUSE_PASSWORD:
+  # CLICKHOUSE_DB:   default
+  #
+  # CLICKHOUSE_SECURE_HOST:     github.demo.trial.altinity.cloud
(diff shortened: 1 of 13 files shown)


filimonov:
Failure on clang: https://bugs.llvm.org/show_bug.cgi?id=16404 (happens only with static linking), looks caused by `__builtin_mul_overflow` call

filimonov:
```
 D:\a\clickhouse-cpp\clickhouse-cpp\ut\columns_ut.cpp(461): error: Expected equality of these values:
  0xffffffffffffffffll
    Which is: -1
  col->At(1)
    Which is: 18446744073709551615
``` 
is a real issue?

Enmk:
> ```
>  D:\a\clickhouse-cpp\clickhouse-cpp\ut\columns_ut.cpp(461): error: Expected equality of these values:
>   0xffffffffffffffffll
>     Which is: -1
>   col->At(1)
>     Which is: 18446744073709551615
> ```
> 
> is a real issue?

Just a broken test, fixed

Enmk (approved):
LGTM

PR #136, opened 2022-01-10
Better tests / CI

filimonov:
preparations for #131 

* for macOS I've added tls offoader proxy, mimicking local clickhouse by tunneling the queries to a cloud instance of clickhouse
* connection setting used in tests now configurable via env variables
* fixed a couple of minor issues in tests, and some platform / compiler specific compilation issues
* changed github actions from workflow to matrix (easier to modify, no hidden dependency from the master branch)
* switch to temporary tables to isolate concurrent test, and simplify access management


.github/workflows/linux.yml
```diff
@@ -26,37 +14,44 @@ jobs:
...
-        - build: gcc-8
-          os: ubuntu-latest
+        - build: gcc-8-ssl
           INSTALL: gcc-8 g++-8
           C_COMPILER: gcc-8
           CXX_COMPILER: g++-8
+          extra_cmake_flags: -DWITH_OPENSSL=ON
+
+        - build: gcc-9-ssl
+          INSTALL: gcc-9 g++-9
```
filimonov, line 39: ups, it look like i didn't intall libssl-dev in every scenario, but somehow it works...

.github/workflows/macos.yml
```diff
@@ -5,42 +5,54 @@ on:
...
-      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DBUILD_TESTS=ON ${{inputs.extra_cmake_flags}}
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DBUILD_TESTS=ON ${{matrix.extra_cmake_flags}}
 
     - name: Build
       run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --target all
 
+    - name: Start tls offoader proxy
+      # that mimics non-secure clickhouse running on localhost
+      # by tunneling queries to remote tls server
+      # (needed because we can't start real clickhouse instance on windows)
```
filimonov, line 46:
```suggestion
```

.github/workflows/linux.yml
Enmk: It could be already installed, but anyway let's make it explicit dependency

Enmk (changes requested):
Minor updates required, mostly comments, please see question inline.

.github/workflows/linux.yml
filimonov: I like your option much more :+1: Wasn't aware that you can make a cartesian product of matrixes this way

PR #139, opened 2022-01-11
Add list of hosts with ports for setting connection

df530:
It is possible to specify multiple hosts to connect to. It approves fault tolerance of connection because when one host isn't alive client can try to connect to another one.
The connection will be set to the first available host.
```cpp
Client client(ClientOptions()
              .SetHost({
                  ClientOptions::HostPort("host1.com", 8000),
                  ClientOptions::HostPort("host2.com"), /// port is ClientOptions.port
              }));
              
```

When `ResetConnection` is called, it tries to connect to the first host. If connection failed, then it tries to connect to the next host in list, while list doesn't end up (round-robin approach). If list ends up, `ResetConnection` throws exception of last host.


PolyProgrammist:
Looks good to me. I think it's better to merge commits in one. And providing more information about the reasons for commit (fault tolerance) and about the implementation (round robin) in pull request description would be great

filimonov:
The idea/motivation is clean. Maybe it would be clearer to push that logic a bit deeper?

Here we have iteration through different IPs which can be extracted from the same name:
https://github.com/ClickHouse/clickhouse-cpp/blob/b36c87fb6f09376f71fbda671cae4c754afc7ea4/clickhouse/base/socket.cpp#L87-L88

Maybe we can generalize that code a bit for multiple hosts/ports cases instead of creating one more retry loop outside?

In that case, connect timeouts should work better.

clickhouse/client.cpp
```diff
@@ -167,6 +190,7 @@ class Client::Impl {
 #endif
 
     ServerInfo server_info_;
+    std::optional<ClientOptions::HostPort> connected_host_port_;
 };
```
Enmk, line 196: Please modify endpoints creation to simplify the `ResetConnection()`
```suggestion
ClientOptions modifyClientOptions(ClientOptions opts)
{
    if (!opts.host.empty())
        opts.endpoints.insert(opts.endpoints.begin(), ClientOptions::Endpoint{opts.host, opts.port});

    return opts;
}

Client::Impl::Impl(const ClientOptions& opts)
    : options_(modifyClientOptions(opts))
{
```

clickhouse/client.cpp
```diff
@@ -296,73 +320,100 @@ void Client::Impl::Ping() {
 }
 
 void Client::Impl::ResetConnection() {
+    connected_host_port_.reset();
+    for (int i = -1; i < int(options_.hosts_ports.size()); ++i) {
+        const ClientOptions::HostPort& host_port = i == -1 ? ClientOptions::HostPort(options_.host) : options_.hosts_ports[i];
```
Enmk, line 325: Please update the list of endpoints before copying it to a `options_` so you don't have to do this. (see above)

clickhouse/client.h
```diff
@@ -196,6 +206,8 @@ class Client {
 
     const ServerInfo& GetServerInfo() const;
 
+    const std::optional<ClientOptions::HostPort>& GetConnectedHostPort() const;
```
Enmk, line 209: Please specify in comments when this function may return `nullopt`.

Enmk (changes requested):
Please add proper unit-tests and fix highlighted issues.

tests/simple/main.cpp
```diff
@@ -510,6 +510,67 @@ int main() {
...
+                                           ClientOptions::HostPort("localhost", 7000), // wrong port
+                                           ClientOptions::HostPort("1127.91.2.1"), // wrong host
+                                           ClientOptions::HostPort("1127.91.2.2"), // wrong host
+                                           ClientOptions::HostPort("notlocalwronghost"), // wrong host
+                                           ClientOptions::HostPort("another_notlocalwronghost"), // wrong host
+                                           correct_host_port,
+                                           ClientOptions::HostPort("localhost", 9001), // wrong port
+                                           ClientOptions::HostPort("1127.911.2.2"), // wrong host
+                                       })
+                              .SetPingBeforeQuery(true));
```
Enmk, line 528: BTW, [this](https://github.com/ClickHouse/clickhouse-cpp/pull/139#discussion_r790474521) allows simplifying client code:
```suggestion
            ClientOptions::Endpoint correct_endpoint{"localhost", 9000};
            Client client(ClientOptions()
                              .SetEndpoints({
                                           {"localhost", 8000}, // wrong port
                                           {"localhost", 7000}, // wrong port
                                           {"1127.91.2.1"}, // wrong host
                                           {"1127.91.2.2"}, // wrong host
                                           {"notlocalwronghost"}, // wrong host
                                           {"another_notlocalwronghost"}, // wrong host
                                           {correct_endpoint},
                                           {"localhost", 9001}, // wrong port
                                           {"1127.911.2.2"}, // wrong host
                                       })
                              .SetPingBeforeQuery(true));
```

Enmk (changes requested):
Please add proper unit-tests and fix highlighted issues.

Enmk:
@DF5HSE ping?

PolyProgrammist:
@Enmk hi, I have just started to continue working on this pr. Hope it will be ready soon

MikhailBurdukov:
Hi! 
We think this feature will be very useful for us, so I would like to complete it. Can we reopen PR or is it better to open a new one?

MikhailBurdukov:
@Enmk ping?

PR #146, opened 2022-01-28
fix: avoid throwing exception in destructor

metahys:
When data compression is enabled,  `CompressedOutput` will be created in `Client::Impl::SendData`. If the socket is unfortunately broken during the execution of `SocketOutput::DoWrite`, a `std::system_error` exception will be thrown. Next, the destructor of `CompressedOutput` will be called during stack unwinding, and then `Flush()` in `~CompressedOutput` will call `SocketOutput::DoWrite` again, eventually causing the program to terminate.

### DIFF
--- a/clickhouse/base/compressed.cpp
+++ b/clickhouse/base/compressed.cpp
@@ -111,7 +111,16 @@ CompressedOutput::CompressedOutput(OutputStream * destination, size_t max_compre
 }
 
 CompressedOutput::~CompressedOutput() {
-    Flush();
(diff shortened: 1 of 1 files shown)


filimonov (approved):

Enmk (changes requested):
IMO it is better to remove Flush() from destructor altogether and make it a caller's responsibility to explicitly call Flush() as a happy-path. In this case all errors that arise are visible to the library user and no need to hide those.

PR #148, opened 2022-02-02
Fixed ColumnIPv4::Append()

Enmk:
Added tests, simplified implementation + few extra minor touches

Closes: #101, #88

### DIFF
--- a/clickhouse/columns/ip4.cpp
+++ b/clickhouse/columns/ip4.cpp
@@ -1,11 +1,7 @@
-
 #include "ip4.h"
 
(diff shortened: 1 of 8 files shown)


filimonov (approved):
LGTM. 

[out of PR scope]: column tests can be improved a bit by introducing real clickhouse roundtrips (to ensure the byte order and other similar things are seen the same way by clickhouse-cpp & clickhouse itself).

Enmk:
> 
We have a 'Client'-tests for that.

PR #152, opened 2022-02-09
OpenSSL updates: allow to configure or skip verification

Enmk:
Added ClientOptions:
  - skip_verification - to skip all verification
  - host_flags - to set hostname verification mode
  - configuration - to configure SSL connection

Updated and refactored tests.
Updated tests/simple/main.cpp to reuse so utility code from unit-tests.

Closes: #142

### DIFF
--- a/clickhouse/base/sslsocket.cpp
+++ b/clickhouse/base/sslsocket.cpp
@@ -27,17 +27,16 @@ std::string getCertificateInfo(X509* cert)
     return std::string(data, len);
 }
 
-void throwSSLError(SSL * ssl, int error, const char * /*location*/, const char * /*statement*/) {
+void throwSSLError(SSL * ssl, int error, const char * /*location*/, const char * /*statement*/, const std::string prefix = "OpenSSL error: ") {
     const auto detail_error = ERR_get_error();
     auto reason = ERR_reason_error_string(detail_error);
     reason = reason ? reason : "Unknown SSL error";
 
     std::string reason_str = reason;
     if (ssl) {
-        // TODO: maybe print certificate only if handshake isn't completed (SSL_get_state(ssl) != TLS_ST_OK)
-        if (auto ssl_session = SSL_get_session(ssl)) {
+        // Print certificate only if handshake isn't completed 
+        if (auto ssl_session = SSL_get_session(ssl); ssl_session && SSL_get_state(ssl) != TLS_ST_OK)
             reason_str += "\nServer certificate: " + getCertificateInfo(SSL_SESSION_get0_peer(ssl_session));
-        }
     }
 
 //    std::cerr << "!!! SSL error at " << location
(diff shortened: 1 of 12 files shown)


clickhouse/base/sslsocket.cpp
```diff
@@ -45,7 +45,42 @@ void throwSSLError(SSL * ssl, int error, const char * /*location*/, const char *
 //              << "\n\t last err: " << ERR_peek_last_error()
 //              << std::endl;
 
-    throw std::runtime_error(std::string("OpenSSL error: ") + std::to_string(error) + " : " + reason_str);
+    throw std::runtime_error((prefix.empty() ? "OpenSSL error: " : prefix) + std::to_string(error) + " : " + reason_str);
```
filimonov, line 48: maybe use "OpenSSL error: " as the default prefix in function signature?

clickhouse/base/sslsocket.cpp
```diff
@@ -143,42 +178,42 @@ SSLSocket::SSLSocket(const NetworkAddress& addr, const SSLParams & ssl_params, S
...
-                    ASN1_STRING_get0_data(addr.get()),
-                    ASN1_STRING_length(addr.get()),
-                    0));
-        } else {
-            HANDLE_SSL_ERROR(ssl, X509_check_host(peer_certificate, hostname.c_str(), hostname.length(), 0, &out_name));
-        }
-    }
+    // Host name verification is done by OpenSSL itself, however if we are connecting to an ip-address,
+    // no verification is made, so we have to do it manually.
+    // Just in case if this is ever required, leave it here commented out.
```
filimonov, line 208: is SSL_set_verify SSL_VERIFY_PEER not needed? 
https://www.openssl.org/docs/man1.1.1/man3/SSL_set_verify.html

at also appears in the examples in https://github.com/iSECPartners/ssl-conservatory/blob/1449f3ae918623699947e24f388c531b30ad2c87/openssl/test_client.c#L115-L119

https://wiki.openssl.org/index.php/Hostname_validation

clickhouse/base/sslsocket.cpp
```diff
@@ -143,42 +178,42 @@ SSLSocket::SSLSocket(const NetworkAddress& addr, const SSLParams & ssl_params, S
     if (!ssl)
         throw std::runtime_error("Failed to create SSL instance");
 
+    std::unique_ptr<ASN1_OCTET_STRING, decltype(&ASN1_OCTET_STRING_free)> ip_addr(a2i_IPADDRESS(addr.Host().c_str()), &ASN1_OCTET_STRING_free);
+
     HANDLE_SSL_ERROR(ssl, SSL_set_fd(ssl, handle_));
     if (ssl_params.use_SNI)
```
filimonov, line 184: sidenote: maybe that is not needed at all, it sounds safe to call SSL_set_tlsext_host_name always.

clickhouse/client.h
```diff
@@ -147,16 +146,45 @@ struct ClientOptions {
...
+         *
+         *  WARNING: settig to true will bypass all SSL session checks, which
+         *  is dangerous, but can be used against self-signed certificates, e.g. for testing purposes.
+         */
+        DECLARE_FIELD(skip_verification, bool, SetSkipVerification, false);
+
+        /** Mode of verifying host ssl certificate against name of the host, set with SSL_set_hostflags.
+         *  For details see https://www.openssl.org/docs/man1.1.1/man3/SSL_set_hostflags.html
+         */
+        DECLARE_FIELD(host_flags, int, SetHostVerifyFlags, DEFAULT_VALUE);
```
filimonov, line 163: Test? For example maybe we can trigger  X509_CHECK_FLAG_NO_WILDCARDS? I guess it should make the `github.demo.trial.altinity.cloud` stop working, because it uses wildcard cert:
```
depth=0 CN = *.demo.trial.altinity.cloud
```

clickhouse/client.h
```diff
@@ -147,16 +146,45 @@ struct ClientOptions {
...
 
+        /** Skip SSL session verification (server's certificate, etc).
+         *
+         *  WARNING: settig to true will bypass all SSL session checks, which
+         *  is dangerous, but can be used against self-signed certificates, e.g. for testing purposes.
+         */
+        DECLARE_FIELD(skip_verification, bool, SetSkipVerification, false);
+
+        /** Mode of verifying host ssl certificate against name of the host, set with SSL_set_hostflags.
+         *  For details see https://www.openssl.org/docs/man1.1.1/man3/SSL_set_hostflags.html
```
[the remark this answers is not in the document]
Enmk: IMO it is ok to refer the official docs, the rest is up to the user.

clickhouse/client.h
Enmk: Rn, it is too hard to set up any complex tests with the tools we have

But I already asked QA team to do the in-depth tests with various tricky ssl (mis)configurations. They'll do it once they have resources.

clickhouse/base/sslsocket.cpp
Enmk: Yep, makes sense to have some safe defaults

PR #157, opened 2022-03-04
"thread safe" client

caiowakamatsuahrefs:
From previous PR #122, advice was taken and now there is a wrapper class where only 1 thread can access any method at any given time. 

There were suggestions to create some wrapper class. So now there is `ThreadSafeClient` (Might need to be renamed to `SingleThreadClient` possibly. 

`ThreadSafeClient` just wraps a regular `Client` and projects all methods with a mutex / lock.

### DIFF
--- a/clickhouse/CMakeLists.txt
+++ b/clickhouse/CMakeLists.txt
@@ -28,7 +28,8 @@ SET ( clickhouse-cpp-lib-src
 
     block.cpp
     client.cpp
-    query.cpp
+        thread_safe_client.cpp
+        query.cpp
 )
 
 IF (WITH_OPENSSL)
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -15,6 +15,7 @@
 #include <vector>
 #include <sstream>
 #include <stdexcept>
+#include <mutex>
 
 #if defined(WITH_OPENSSL)
 #include "base/sslsocket.h"
@@ -162,6 +163,8 @@ private:
 
     std::unique_ptr<Socket> socket_;
 
+    std::recursive_mutex stream_mutex_;
+
 #if defined(WITH_OPENSSL)
     std::unique_ptr<SSLContext> ssl_context_;
 #endif
@@ -198,6 +201,8 @@ Client::Impl::~Impl()
 { }
 
 void Client::Impl::ExecuteQuery(Query query) {
+    auto lock = std::unique_lock(stream_mutex_);
+
     EnsureNull en(static_cast<QueryEvents*>(&query), &events_);
 
     if (options_.ping_before_query) {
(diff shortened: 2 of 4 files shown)


Enmk (changes requested):
Thank you for a PR, but current solution has many drawbacks: having to sync the API of the `ThreadSafeClient` with `Client`, recursive mutex, no tests.

TBH, I was expecting something like a generic wrapper:
```cpp
template <typename T, typename Mutex = std::mutex, typename Lock = std::lock_guard<Mutex> >
class TreadSafeWrapper {
public:
    using MutexType = Mutex;
    using LockType = Lock;

private:
    std::unique_ptr<T> synchronized_;
    MutexType mutex;

public:
    template <typename ...Args>
    TreadSafeWrapper(Args && ... args)
        : synchronized_(std::make_unique<T>(std::forward<Args>(args)...))
    {}

    TreadSafeWrapper(std::unique_ptr<T> synchronized)
        : synchronized_(std::move(synchronized))
    {}

    template <typename U>
    class AccessWrapper {
        U & object_;
        LockType lock_;
    public:
        AccessWrapper(U & object, MutexType & mutex)
            : object_(object),
              lock_(mutex)
        {}

        T * operator->() {
            return &object_;
        }

        const T * operator->() const {
            return &object_;
        }
    };

    auto operator->() const {
        return AccessWrapper<const T>(*synchronized_, mutex);
    }

    auto operator->() {
        return AccessWrapper<T>(*synchronized_, mutex);
    }
};
```

So please:
- consider introducing a wrapper (or finding an open-source one we can use as a third-party dependency).
- add some tests for concurent use of `ThreadSafeClient`

Enmk (changes requested):
Thank you for a PR, what are supported use-cases? Sending a query and canceling it?

Please add some tests of concurrent use of `ThreadSafeWrapper`.
Also, the current solution has a minor drawbacks:
- recursive_mutex
- API of `Client` and `ThreadSafeWrapper` has to be manually synchronized.
Please consider some generic wrapper instead of `ThreadSafeWrapper`:

```cpp
template <typename T, typename Mutex = std::mutex, typename Lock = std::lock_guard<Mutex> >
class TreadSafeWrapper {
public:
    using MutexType = Mutex;
    using LockType = Lock;

private:
    std::unique_ptr<T> synchronized_;
    MutexType mutex;

public:
    template <typename ...Args>
    TreadSafeWrapper(Args && ... args)
        : synchronized_(std::make_unique<T>(std::forward<Args>(args)...))
    {}

    TreadSafeWrapper(std::unique_ptr<T> synchronized)
        : synchronized_(std::move(synchronized))
    {}

    template <typename U>
    class AccessWrapper {
        U & object_;
        LockType lock_;
    public:
        AccessWrapper(U & object, MutexType & mutex)
            : object_(object),
              lock_(mutex)
        {}

        T * operator->() {
            return &object_;
        }

        const T * operator->() const {
            return &object_;
        }
    };

    auto operator->() const {
        return AccessWrapper<const T>(*synchronized_, mutex);
    }

    auto operator->() {
        return AccessWrapper<T>(*synchronized_, mutex);
    }
};
```

Or you can find some header-only library that does exactly that and use it as third-party.

Enmk:
Closed due to inactivity

PR #164, opened 2022-03-15
introduce exceptions hierarchy

### LINKED ISSUE #145: Consider incorporating exceptions into type system
As for now there are 62 places where `std::runtime_error` is thrown and 9 places where `std::system_error` is thrown,
and this makes it really hard to distinguish different kinds of errors programmatically.

Would be great if you could introduce something like `ParsingError`, `NetworkError` etc.

itrofimow:
Closes #145 


itrofimow:
Hi! 
I'll have a fresh look tomorrow, probably fix a couple of places, and then mark it as ready

itrofimow:
I am not certain about what should be thrown instead of base `Error` in some places, but other than that i believe it's ready for review.

I didn't find any test for throwing logic, thus i didn't add any

@Enmk

clickhouse/base/wire_format.cpp
```diff
@@ -38,7 +40,7 @@ void WireFormat::WriteAll(OutputStream& output, const void* buf, size_t len) {
     }
 
     if (len) {
-        throw std::runtime_error("Failed to write " + std::to_string(original_len)
+        throw Error("Failed to write " + std::to_string(original_len)
```
itrofimow, line 43: Not sure what should be thrown here

clickhouse/client.cpp
```diff
@@ -489,12 +485,12 @@ bool Client::Impl::ReadBlock(InputStream& input, Block* block) {
 
         if (ColumnRef col = CreateColumnByType(type, create_column_settings)) {
             if (num_rows && !col->Load(&input, num_rows)) {
-                throw std::runtime_error("can't load column '" + name + "' of type " + type);
+                throw Error("can't load column '" + name + "' of type " + type);
```
itrofimow, line 488: Not sure what should be thrown here

clickhouse/columns/lowcardinality.cpp
```diff
@@ -197,42 +197,42 @@ auto Load(ColumnRef new_dictionary_column, InputStream& input, size_t rows) {
...
+        throw ValidationError("HasAdditionalKeysBit is missing.");
 
     uint64_t number_of_keys;
     if (!WireFormat::ReadFixed(input, &number_of_keys))
-        throw std::runtime_error("Failed to read number of rows in dictionary column.");
+        throw ProtocolError("Failed to read number of rows in dictionary column.");
 
     if (!new_dictionary_column->Load(&input, number_of_keys))
-        throw std::runtime_error("Failed to read values of dictionary column.");
+        throw Error("Failed to read values of dictionary column.");
```
itrofimow, line 228: Not sure what should be thrown here

PR #166, opened 2022-03-25
make ItemView compatible with libcxx

itrofimow:
In libcxx version we currently use this constructor is deleted
`basic_string_view(nullptr_t, size_t) = delete;`

As of c++23 nullptr constructor is also deleted, and default constructor already does [just that](https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view): `.data = nullptr, .size = 0`

### DIFF
--- a/clickhouse/columns/itemview.h
+++ b/clickhouse/columns/itemview.h
@@ -43,7 +43,7 @@ public:
     }
 
     explicit ItemView()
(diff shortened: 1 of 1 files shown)


itrofimow:
Closed in favor of https://github.com/ClickHouse/clickhouse-cpp/pull/167 for checks restart ¯\_(ツ)_/¯

PR #167, opened 2022-03-25
replace explicit string_view initialization from {nullptr, 0} with default ctor

itrofimow:


Enmk:
Hi! Thank you for contribution, could you please specify a libcxx version and error code+message?

itrofimow:
Well, turns out we use a forked version on llvms libcxx with `(nullptr, size_t)` constructor explicitly deleted, my bad.

However point about explicit invocation still stand - default ctor of string_view already does just that

PR #171, opened 2022-04-23
Optimize array access

Enmk:
`ColumnArrayT<X>` template that simplifies and optimizes both reading and appending data to Array.
* For more optimal access to the Array's data, both `ColumnArrayT<X>::operator[]` and `ColumnArrayT<X>::At()` provide a wrapper-value rather that a full-blown column. So from underlying column is not copied, but rather accessed via read-only "view".
* `ColumnArrayT<X>::Append()` overloads work for anything iteratable. Users don't have to wrap values into `ColumnX` anymore.
* For convenience, values obtained from both `ColumnArrayT<X>::operator[]` and `ColumnArrayT<X>::At()` provide type-safe access to the typed data.
* Existing `ColumnArray` instance can be type-safe wrapped into appropriate `ColumnArrayT<NestedColumnType>` so users can benefit from features above for columns obtained from `Block`.

## Examples 
* `Array(UInt64)`:
```
auto array = std::make_shared<clickhouse::ColumnArrayT<ColumnUInt64>>();
array->Append({0, 1, 2});

EXPECT_EQ(0u, array->At(0).At(0)); // 0
EXPECT_EQ(1u, (*array)[0][1]);     // 1
```

* `Array(FixedString(6))`:
```
using namespace std::literals;
auto array = std::make_shared<clickhouse::ColumnArrayT<ColumnFixedString>>(6); // argument is passed to ColumnFixedString constructor
array->Append({"hello", "world"});

// Additional \0 since strings are padded from right with zeros in FixedString(6).
EXPECT_EQ("hello\0"sv, array->At(0).At(0));

auto row = array->At(0);
EXPECT_EQ("hello\0"sv, row.At(0));
EXPECT_EQ(6u, row[0].length());
EXPECT_EQ("hello", row[0].substr(0, 5));

EXPECT_EQ("world\0"sv, (*array)[0][1]);
```

* `Array(Array(UInt64))`
```
// Nested 2D-arrays are supported too:
auto array = std::make_shared<clickhouse::ColumnArrayT<clickhouse::ColumnArrayT<ColumnUInt64>>>();
array->Append(std::vector<std::vector<unsigned int>>{{0}, {1, 1}, {2, 2, 2}});

EXPECT_EQ(0u, array->At(0).At(0).At(0)); // 0
EXPECT_EQ(1u, (*array)[0][1][1]);        // 1
```

See unit-tests for more examples.


clickhouse/columns/array.h
```diff
@@ -38,19 +53,209 @@ class ColumnArray : public Column {
...
+    static auto Wrap(Column&& col) {
+        return Wrap(std::move(dynamic_cast<ColumnArray&&>(col)));
+    }
+
+    // Helper to simplify integration with other APIs
+    static auto Wrap(ColumnRef&& col) {
+        return Wrap(std::move(*col->AsStrict<ColumnArray>()));
+    }
+
+    class ArrayWrapper {
```
Enmk, line 130: `ArrayView` ?

ut/column_array_ut.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+    EXPECT_TRUE(CompareRecursive(values, array));
+}
+
+TEST(ColumnsCase, ArrayTArrayTWrap) {
+    // TODO(nemkov): wrap 2D array
+    // Check that ColumnArrayT can wrap a pre-existing ColumnArray,
+    // pre-existing data is kept intact and new rows can be inserted.
+
+    const std::vector<std::vector<std::vector<uint64_t>>> values = {
```
Enmk, line 209: uncomment values

clickhouse/columns/array.cpp
```diff
@@ -5,43 +5,58 @@
...
 }
 
+// Explicitly sharing `data_` column is required by ColumnArrayT
+ColumnArray::ColumnArray(ColumnArray&& other)
+    : Column(other.Type())
+    , data_(std::move(other.data_))
+    , offsets_(std::move(other.offsets_))
+{
+}
+
```
arthurpassos, line 28: What do you think about leaving a comment mentioning that these exist because of the ColumnArrayT?

clickhouse/columns/array.h
arthurpassos: I think that would be better.

clickhouse/columns/array.h
```diff
@@ -38,19 +53,209 @@ class ColumnArray : public Column {
...
+        const size_t offset_;
+        const size_t size_;
+
+    public:
+        using ValueType = typename NestedColumnType::ValueType;
+
+        ArrayWrapper(std::shared_ptr<NestedColumnType> data, size_t offset = 0, size_t size = std::numeric_limits<size_t>::max())
+            : typed_nested_data_(data)
+            , offset_(offset)
+            , size_(std::min(typed_nested_data_->Size() - offset, size))
```
arthurpassos, line 141: If `size` is bigger than `typed_nested_data_->Size() - offset`, I assume the client would expect to get `typed_nested_data_->Size() - offset` of size. If that's the case, then it's fine. Otherwise, maybe we should throw an exception.

clickhouse/columns/array.h
```diff
@@ -38,19 +53,209 @@ class ColumnArray : public Column {
...
+        {}
+
+        inline auto operator[](size_t index) const {
+            return (*typed_nested_data_)[offset_ + index];
+        }
+
+        inline auto At(size_t index) const {
+            if (index >= size_)
```
arthurpassos, line 149: I might be missing something, but at a first glance I feel like this should be checking for `index + offset` and not only `index`. Plus, shouldn't this check also be present on the `[] operator`?

arthurpassos (commented):
I still need to go over some details and the tests. Submeting partial review so we can work in parallel.

ut/client_ut.cpp
```diff
@@ -913,6 +915,168 @@ TEST_P(ClientCase, Query_ID) {
...
+
+    {
+        // ([[]])
+        auto row = result->GetAsColumnTyped<ColumnArray>(2);
+        ASSERT_EQ(1u, row->Size());
+        EXPECT_TRUE(CompareRecursive(std::vector<uint64_t>{}, *row->GetAsColumnTyped<ColumnUInt64>(0)));
+    }
+}
+
+ColumnRef RoundtripColumnValues(Client& client, ColumnRef expected) {
```
arthurpassos, line 992: What do you think about renaming this to something like `InsertAndQuery`? I understand the reasoning behind Roundtrip, but it doesn't seem to be the most intuitive one.

clickhouse/columns/array.h
```diff
@@ -38,19 +53,209 @@ class ColumnArray : public Column {
...
+     *  MUST NOT BE USED IN ANY WAY, it is only safe to dispose it.
+     *
+     *  Throws an exception of `col` is of wrong type, it is safe to use original col in this case.
+     *  This is a static method to make such conversion verbose.
+     */
+    static auto Wrap(ColumnArray&& col) {
+        if constexpr (std::is_base_of_v<ColumnArray, NestedColumnType> && !std::is_same_v<ColumnArray, NestedColumnType>) {
+            // assuming NestedColumnType is ArrayT specialization
+
+            auto result = std::make_shared<ColumnArrayT<NestedColumnType>>(NestedColumnType::Wrap(col.GetData()));
```
arthurpassos, line 112: What if Array`<Array<Array<Array<Array<Array<Array<Array<Array<Array<Array<Array<int>>>>>>>>>>>>`.

Just kidding :P

Enmk: Well... I would hate to write that kind of test. Let's hope that no clients have such use cases...

clickhouse/columns/array.h
Enmk: Well, trying to follow STL's lead here: `operator[]` is as simple as it could be: no checks, no safety, whatever bad happens - is a user's fault.

clickhouse/columns/array.h
Enmk: well, we build `ArrayWrapper` ourselves, and IMO such situation would never happen.

clickhouse/columns/array.h
```diff
@@ -38,19 +53,209 @@ class ColumnArray : public Column {
...
+    /** Create a ColumnArrayT from a ColumnArray, without copying data and offsets.
+     *  Ownership of column internals is transferred to returned object, original (argument) object
+     *  MUST NOT BE USED IN ANY WAY, it is only safe to dispose it.
+     *
+     *  Throws an exception of `col` is of wrong type, it is safe to use original col in this case.
+     *  This is a static method to make such conversion verbose.
+     */
+    static auto Wrap(ColumnArray&& col) {
+        if constexpr (std::is_base_of_v<ColumnArray, NestedColumnType> && !std::is_same_v<ColumnArray, NestedColumnType>) {
+            // assuming NestedColumnType is ArrayT specialization
```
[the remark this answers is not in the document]
Enmk: Perhaps not so important here

clickhouse/columns/array.cpp
```diff
@@ -5,43 +5,58 @@
 namespace clickhouse {
 
 ColumnArray::ColumnArray(ColumnRef data)
+    : Column(Type::CreateArray(data->Type()))
+    , data_(data->CloneEmpty())
```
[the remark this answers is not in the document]
Enmk: Removed that, since it breaks previous user code and complicates implementation of other methods. Just added a warning comment that users shouldn't abuse this constructor.

clickhouse/columns/array.h
Enmk: Fixed

arthurpassos (approved):

Issue #140, opened 2022-01-19
Exception on inserting intto Array(LowCardinality(String))


typeryougishiki:
Code like this:
```client.Execute("CREATE TABLE IF NOT EXISTS test.array_locard_test(arr Array(LowCardinality(String))) Engine = Memory");
auto arr_locard_string = make_shared<ColumnArray>(make_shared<ColumnLowCardinalityT<ColumnString>>());
auto locard_string = make_shared<ColumnLowCardinalityT<ColumnString>>();
locard_string->Append("aaa");
locard_string->Append("bbb");
arr_locard_string->AppendAsColumn(locard_string);
Block block;
block.AppendColumn("arr", arr_locard_string);
client.Insert("test.array_locard_test",block);
```
And exception is "DB::Exception: Invalid version for SerializationLowCardinality key column."

Enmk:
Hi @typeryougishiki! Thanks for the report, going to fix it.

haoxiaokun:
when fix this bug?

arthurpassos:
Working on it

Enmk:
Closed with https://github.com/ClickHouse/clickhouse-cpp/pull/173

PR #173, opened 2022-04-26
Fix Array of low cardinality types serialization/deserialization

arthurpassos:
## Context

Array of LowCardinality serialization/deserialization was broken as reported in https://github.com/ClickHouse/clickhouse-cpp/issues/140. Issue was that clickhouse-cpp was serializing/deserializing the array offsets before the lowcardinality key_version and Clickhouse was expecting it to come after the key_version.

Solution was to implement a prefix, body and suffix serialization/deserialization approach similar to what can be found in Clickhouse/NativeWriter.

## Changes
1. Add SavePrefix, SaveBody and SaveSuffix methods to Column base class.
2. Add and implement Save template method in Column base class. Column::Save calls SavePrefix, SaveBody and SaveSuffix respectively.
3. Add unit and integration tests for array of low cardinality. See ut/array_of_low_cardinality_tests.cpp

## Testing
1. Ran existing tests and added new ones.

## Closes
*  https://github.com/ClickHouse/clickhouse-cpp/issues/140


clickhouse/columns/column.cpp
```diff
@@ -0,0 +1,38 @@
...
+    /// does nothing by default
+}
+
+void Column::SaveSuffix([[maybe_unused]] OutputStream* output) {
+    /// does nothing by default
+}
+
+/// Saves column data to output stream.
+void Column::Save(OutputStream* output) {
+    SavePrefix(output);
```
Enmk, line 33: Please add a comment inside a function body explaining why we have to split saving in 3 steps. The best comment would also explain why this is done in CH, not only stating something like "follow CH serialization schema"

clickhouse/columns/column.h
```diff
@@ -42,11 +42,28 @@ class Column : public std::enable_shared_from_this<Column> {
...
+    virtual bool LoadBody(InputStream* input, size_t rows);
+
+    /// Saves column prefix to output stream. Column types with prefixes must implement it.
+    virtual void SavePrefix(OutputStream* output);
+
+    /// Saves column body to output stream.
+    virtual void SaveBody(OutputStream* output);
+
+    /// Saves column suffix to output stream. Column types with suffixes must implement it.
+    virtual void SaveSuffix(OutputStream* output);
```
Enmk, line 62: It looks like no column has meaningful implementation of `SaveSuffix`, except ones calling on `SaveSuffix` on nested column(s). So we may want to remove `SaveSuffix` altogether.

clickhouse/columns/column.h
```diff
@@ -42,11 +42,28 @@ class Column : public std::enable_shared_from_this<Column> {
...
 
-    /// Saves column data to output stream.
-    virtual void Save(OutputStream* output) = 0;
+    /// Template method to save to output stream. It'll call SavePrefix, SaveBody and SaveSuffix respectively
+    /// Should be called only once from the client. Derived classes should not call it.
+    virtual void Save(OutputStream* output);
```
Enmk, line 66: Why leaving `Save` a virtual method? Any of the derived columns may want to change serialization order?

ut/array_of_low_cardinality_tests.cpp
```diff
@@ -0,0 +1,99 @@
...
+        { "cc" }
+    };
+
+    auto column = buildTestColumn(testData);
+
+    Block block;
+    block.AppendColumn("arr", column);
+
+    client.Execute("DROP TABLE test.array_lc");
+    client.Execute("CREATE TABLE IF NOT EXISTS test.array_lc (arr Array(LowCardinality(String))) ENGINE = Memory");
```
Enmk, line 84: It must be a temporary table due to CI/CD quirks on some platforms. E.g. on Windows (and MacOS?) we relay queries via local proxy to a remote server, and on that server we don't have permissions to create any non-temporary tables.

Enmk (changes requested):
Good, but not yet ready to merge

clickhouse/columns/array.h
```diff
@@ -24,11 +24,20 @@ class ColumnArray : public Column {
...
     /// Loads column data from input stream.
-    bool Load(InputStream* input, size_t rows) override;
+    bool LoadBody(InputStream* input, size_t rows) override;
+
+    /// Saves column prefix to output stream.
+    void SavePrefix(OutputStream* output) override;
 
     /// Saves column data to output stream.
-    void Save(OutputStream* output) override;
+    void SaveBody(OutputStream* output) override;
```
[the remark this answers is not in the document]
arthurpassos: Yeah, thought of that as an option. Fine for me.

clickhouse/columns/column.cpp
```diff
@@ -0,0 +1,38 @@
+#include "column.h"
+
+namespace clickhouse {
+
+bool Column::LoadPrefix([[maybe_unused]] InputStream* input, [[maybe_unused]] size_t rows) {
```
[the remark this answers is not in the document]
arthurpassos: 👍

clickhouse/columns/column.cpp
```diff
@@ -0,0 +1,38 @@
...
+    return true;
+}
+
+bool Column::LoadBody([[maybe_unused]] InputStream* input, [[maybe_unused]] size_t rows) {
+    /// does nothing by default
+    return true;
+}
+
+bool Column::Load(InputStream* input, size_t rows) {
+    return LoadPrefix(input, rows) && LoadBody(input, rows);
```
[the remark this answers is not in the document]
arthurpassos: As you can see in [Clickhouse](https://github.com/ClickHouse/ClickHouse/blob/master/src/Formats/NativeReader.cpp#L78), it only calls `prefix` and `body`. I was just trying to be consistent, but I do agree that it's weird. As you mentioned in a comment below, so far we have no useful implementation of `SaveSuffix`, so my suggestion is that we remove it and keep this one as it is.

clickhouse/columns/column.cpp
arthurpassos: Ok. I don't know all the reasons it was split in three parts, but I'll mention the Array(LowCardinality) example.

clickhouse/columns/column.h
arthurpassos: 👍

clickhouse/columns/column.h
arthurpassos: Will make it non virtual

ut/array_of_low_cardinality_tests.cpp
arthurpassos: 👍

Enmk (approved):
LGTM

huyphams:
My app is still crashing after this merge, let me see which is the issue.

huyphams:
I found the issue, it will not work if the array of Array(LowCardinality(String)) is empty. For example:

 const auto testData = std::vector<std::vector<std::string>> {
        { "aa", "bb" },
        { "cc" },
        {}
    };

PR #196, opened 2022-07-07
fix #195

### LINKED ISSUE #195: windows connection database crash
I bet no one in the real world has used clickhouse-cpp on windows, not even for testing.

Using the following code on windows results in an error code: `WSAHOST_NOT_FOUND`
https://github.com/ClickHouse/clickhouse-cpp/blob/de70d975a19655e64d6d3037dd1402be02442c08/clickhouse/base/socket.cpp#L194

So it will crash here
https://github.com/ClickHouse/clickhouse-cpp/blob/de70d975a19655e64d6d3037dd1402be02442c08/clickhouse/base/socket.cpp#L200

we need
```
#if defined(_unix_)
    if (!Singleton<LocalNames>()->IsLocalName(host)) {
        // https://linux.die.net/man/3/getaddrinfo
        // If hints.ai_flags includes the AI_ADDRCONFIG flag,
        // then IPv4 addresses are returned in the list pointed to by res only
        // if the local system has at least one IPv4 address configured,
        // and IPv6 addresses are only returned if the local system
        // has at least one IPv6 address configured.
        // The loopback address is not considered for this case
        // as valid as a configured address.
        hints.ai_flags |= AI_ADDRCONFIG;
    }
#endif
```

At the same time, I don't know why windows can never receive database messages here,
https://github.com/ClickHouse/clickhouse-cpp/blob/de70d975a19655e64d6d3037dd1402be02442c08/clickhouse/base/socket.cpp#L335

so inevitably it crashes in the code behind
https://github.com/ClickHouse/clickhouse-cpp/blob/de70d975a19655e64d6d3037dd1402be02442c08/clickhouse/base/socket.cpp#L342

By the way I am connecting to the remote ip

inaryart:
Maybe need to add a macro

closes: #195

### DIFF
--- a/clickhouse/base/socket.cpp
+++ b/clickhouse/base/socket.cpp
@@ -182,6 +182,7 @@ NetworkAddress::NetworkAddress(const std::string& host, const std::string& port)
     hints.ai_family = PF_UNSPEC;
     hints.ai_socktype = SOCK_STREAM;
 
+#if defined(_unix_) 
     if (!Singleton<LocalNames>()->IsLocalName(host)) {
         // https://linux.die.net/man/3/getaddrinfo
         // If hints.ai_flags includes the AI_ADDRCONFIG flag,
@@ -193,6 +194,7 @@ NetworkAddress::NetworkAddress(const std::string& host, const std::string& port)
         // as valid as a configured address.
         hints.ai_flags |= AI_ADDRCONFIG;
     }
+#endif
 
     const int error = getaddrinfo(host.c_str(), port.c_str(), &hints, &info_);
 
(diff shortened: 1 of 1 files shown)


clickhouse/base/socket.cpp
```diff
@@ -182,6 +182,7 @@ NetworkAddress::NetworkAddress(const std::string& host, const std::string& port)
     hints.ai_family = PF_UNSPEC;
     hints.ai_socktype = SOCK_STREAM;
 
+#if defined(_unix_)
```
Enmk, line 185: Please add some context to the fix in comments, like adding `AI_ADDRCONFIG` on Windows causes `getaddrinfo` to fail with `WSAHOST_NOT_FOUND` on host addr "".

Enmk (changes requested):
Please add some background info to the PR: why it fails, maybe links to the corresponding docs? 
Also, please address comments.

PR #200, opened 2022-07-21
Avoid memory copy in ColumnString and Add Reserve function to reduce memory allocation times

King-Authur:
Avoid memory copy in ColumnString and Add Reserve function to reduce memory allocation times.


King-Authur:
In the working environment, we encountered the situation that the clickhouse machine hangs up several times. At this time, the server can receive the client's request, but will not respond, which will cause the client to be stuck in the recv() link, which will affect the subsequent data. write.
Adding the timeout config and cooperating with the logic of switching the write replica when the write fails can avoid the above problems.

Enmk:
Hi! Thank you for your contribution, there are some issues with this PR though.

There are 3 unrelated features, please split it into 3 PRs:
- socket timeout
- ColumnString optimizations
- Reserve

Please add some tests that cover ColumnString and Reserve.
Also, maybe consider making `Reserve` a virtual method of `Column` and providing reasonable defaults to other columns (Decimal, Enum, Tuple, etc) too.

Enmk:
Seems to be inactive, closing. Please feel free to re-open, update, and address highlighted issues.

PR #205, opened 2022-07-30
Add recv/send timeouts to socket

horoshenkiy:
When I was working with the library, I found the following behavior: when my application is connecting to the clickhouse server, it may stick forever due to connection problems.

A connection to the server can be established, but nothing can be read from the socket because you try to connect to another running application in your system. 

For resolving this problem I added receive and send timeouts to socket.

### DIFF
--- a/clickhouse/base/socket.cpp
+++ b/clickhouse/base/socket.cpp
@@ -112,6 +112,16 @@ void SetNonBlock(SOCKET fd, bool value) {
 #endif
 }
 
+void SetTimeout(SOCKET fd, const SocketTimeoutParams& timeout_params) {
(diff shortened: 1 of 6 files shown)


belousotroll:
Hello! We faced this problem in our project and this PR is the solution. Do you need any help to merge?

horoshenkiy:
> Hello! We faced this problem in our project and this PR is the solution. Do you need any help to merge?

Yes, I don’t have rights for merging

belousotroll:
@horoshenkiy I don't have rights either. I mean do you know anything I can help with? PR is 24 days old. Or the only way is to wait for a contributor?

PR #210, opened 2022-08-14
fix CreateColumnFromAst

1261385937:
`case TypeAst::Enum` in `CreateColumnFromAst `  has bug when `ast.elements.size() ==1`


Enmk (changes requested):
Thank you for the contribution, but could you please provide a test?

1261385937:
It is ok

PR #212, opened 2022-08-26
optimize for ColumnString, move the column data if possible, without memory copy.

1261385937:


Enmk:
Hi @1261385937 ! Please address failing builds/tests

PR #216, opened 2022-09-15
fix windows compile error (C-style designated initializer syntax)

1261385937:


1261385937:
This will block 6 hours, because of no windows socket timeout implementation.<br>
Close this pr

PR #217, opened 2022-09-15
fix pr #205

1261385937:


1261385937:
@Enmk , I close the #212.<br>
After #217 merged, I will resubmit #212

clickhouse/base/socket.h
```diff
@@ -83,13 +83,16 @@ class SocketFactory {
 
 
 struct SocketTimeoutParams {
-    const std::chrono::seconds recv_timeout {0};
-    const std::chrono::seconds send_timeout {0};
+    const std::chrono::seconds recv_timeout_s {0};
+    const std::chrono::seconds send_timeout_s {0};
+    const std::chrono::microseconds recv_timeout_us{ 0 };
```
Enmk, line 88: Why do we need separate microseconds and seconds timeout? Looks like WIN API implementation detail creeps into library interface, which is not really cool.

1261385937:
@Enmk  It is ok now

PR #219, opened 2022-10-01
Separate building static and shared libs

Jihadist:
Introduce separating building static and shared libs, depends on cmake out-of-the-box option BUILD_SHARED_LIBS https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html. Also creates alias target for transparently linking to static or shared lib.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ INCLUDE (cmake/openssl.cmake)
 
 OPTION (BUILD_BENCHMARK "Build benchmark" OFF)
 OPTION (BUILD_TESTS "Build tests" OFF)
+OPTION (BUILD_SHARED_LIBS "Build shared libs" OFF)
 OPTION (WITH_OPENSSL "Use OpenSSL for TLS connections" OFF)
 
 PROJECT (CLICKHOUSE-CLIENT)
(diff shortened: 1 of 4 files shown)


Enmk (changes requested):
Hi, @Jihadist, thank you for contribution! Few notes:

- Please provide motivation for the change in PR description.
- It is not clear which version is now used in CI/CD testing, could you please add job to test both static and shared build?
- Any reason (besides final binary file name) to keep both targets:`clickhouse-cpp-lib-static` and `clickhouse-cpp-lib`? From my perspective, only one is going to be built every time. 

IMO, for static builds only, we'd better add an alias: `clickhouse-cpp-lib-static` (and also make sure that output binary name [stays `clickhouse-cpp-lib-static`](https://stackoverflow.com/a/45663205). Everywhere else just use a `clickhouse-cpp-lib`. This is not to break user's code.

Jihadist:
Ok, @Enmk 
- Reason: i wanna provide a conan recipe to conan-center-index for clickhouse-cpp. Conan does not support static and shared libs at the same time. Library should be build as static or as shared. 
- Ok, i'll try to figure out how to do that and change pr
- If project uses library with cmake FetchContent or as subdirectory project may depend on static target name. If we change target name for static libs it will be cmake error. There is only one reason why I do not change target name for static.

Enmk:
Regarding the CI/CD testing: IMO you might want to add another matrix parameter, [like here](https://github.com/ClickHouse/clickhouse-cpp/blob/master/.github/workflows/linux.yml#L17).
Something like:
```diff
      matrix:
        compiler: [clang-6, gcc-7, gcc-8, gcc-9]
        ssl: [ssl_ON, ssl_OFF]
+       build_type: [static, shared]
        include:
...
+       - build_type: shared
+         CMAKE_BUILD_TYPE_FLAG: -BUILD_SHARED_LIBS=ON
```

etc.

Enmk:
Also, if you wish to add that as a Conan module, please free to add any related CI/CD workflows

Jihadist:
> Regarding the CI/CD testing: IMO you might want to add another matrix parameter,

I did this but looks like windows shared build is broken

Jihadist:
If someone can help with windows shared build with export all, it will be nice

1261385937:
> If someone can help with windows shared build with export all, it will be nice

Addressed in #226

1261385937:
> If someone can help with windows shared build with export all, it will be nice

Addressed in #226

Enmk:
Another attempt to fix windows build here: #227 227

Enmk (approved):
LGTM

Issue #214, opened 2022-09-08
Error in running example


wzz1234wzz:
When compiling the examples in the warehouse in centos7 environment, the following error is reported：

In file included from /opt/thirdpart/ck/include/clickhouse/columns/column.h:3,
                 from /opt/thirdpart/ck/include/clickhouse/block.h:3,
                 from /opt/thirdpart/ck/include/clickhouse/query.h:3,
                 from /opt/thirdpart/ck/include/clickhouse/client.h:3,
                 from /home/centos/ckTest/src/main.cpp:1:
/opt/thirdpart/ck/include/clickhouse/columns/../types/types.h:215:10: error: 'string_view' in namespace 'std' does not name a type
  215 |     std::string_view GetEnumName(int16_t value) const;
      |          ^~~~~~~~~~~
/opt/thirdpart/ck/include/clickhouse/columns/../types/types.h:215:5: note: 'std::string_view' is only available from C++17 onwards
  215 |     std::string_view GetEnumName(int16_t value) const;
      |     ^~~
/opt/thirdpart/ck/include/clickhouse/columns/../types/types.h:221:56: error: 'string_view' is not a member of 'std'
  221 |     using ValueToNameType     = std::map<int16_t, std::string_view>;
      |                                                        ^~~~~~~~~~~
/opt/thirdpart/ck/include/clickhouse/columns/../types/types.h:221:56: note: 'std::string_view' is only available from C++17 onwards
/opt/thirdpart/ck/include/clickhouse/columns/../types/types.h:221:56: error: 'string_view' is not a member of 'std'
/opt/thirdpart/ck/include/clickhouse/columns/../types/types.h:221:56: note: 'std::string_view' is only available from C++17 onwards
/opt/thirdpart/ck/include/clickhouse/columns/../types/types.h:221:67: error: template argument 2 is invalid
  221 |     using ValueToNameType     = std::map<int16_t, std::string_view>;
      |                                                                   ^
/opt/thirdpart/ck/include/clickhouse/columns/../types/types.h:221:67: error: template argument 4 is invalid
/opt/thirdpart/ck/include/clickhouse/columns/../types/types.h:223:33: error: 'ValueToNameType' does not name a type
  223 |     using ValueToNameIterator = ValueToNameType::const_iterator;
      |                                 ^~~~~~~~~~~~~~~
/opt/thirdpart/ck/include/clickhouse/columns/../types/types.h:225:5: error: 'ValueToNameType' does not name a type
[...]

PR #220, opened 2022-10-05
Correct example codes

peter279k:
# Changed log

- This issue is to correct the example codes and it's described in this [comment](https://github.com/ClickHouse/clickhouse-cpp/issues/214#issuecomment-1268116484) clearly.


filimonov (approved):
LGTM

PR #221, opened 2022-10-09
Support client write info 

den818:
I want to implement supporting query parameters. It requires raising the version of the protocol. The gap between the current server version of the protocol and the current client library version of the protocol is huge for one PR. So I decided to create several small PR and raise versions step by step.

### DIFF
--- a/clickhouse/block.cpp
+++ b/clickhouse/block.cpp
@@ -71,6 +71,11 @@ const BlockInfo& Block::Info() const {
     return info_;
 }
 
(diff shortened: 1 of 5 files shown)


Enmk (changes requested):
Thank you for contribution! Please add some test that verify that `written_rows` and `written_bytes` are non-zero with apprpriate query.

den818:
Server support CLIENT_WRITE_INFO since  version 19.9

PR #226, opened 2022-10-16
fix win compile error if enable CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS

1261385937:


clickhouse/query.h
```diff
@@ -116,7 +116,7 @@ class Query : public QueryEvents {
         return *this;
     }
 
-    static const std::string default_query_id;
+    inline static const std::string default_query_id{};
```
Jihadist, line 119: Can it be constexpr?

1261385937: Hi, pretty default_query_id

clickhouse/columns/string.h
```diff
@@ -19,14 +19,10 @@ class ColumnFixedString : public Column {
 
     explicit ColumnFixedString(size_t n);
 
-    template <typename Values>
-    ColumnFixedString(size_t n, const Values & values)
```
Enmk, line 23: Removing this constructor breaks client code:
```
        const auto values = std::array<const char*, 2>{
                "first row",
                "second row"};
        ColumnFixedString(16, values);
```

Enmk (changes requested):
Please provide compiler error message with some background in PR description, RN it is not entirely clear why this change required.

1261385937:
another provided

clickhouse/columns/string.h
1261385937: just support std::vector<std::string> or std::vector<std::string_view> may be enough

1261385937: > Removing this constructor breaks client code:
> 
> ```
>         const auto values = std::array<const char*, 2>{
>                 "first row",
>                 "second row"};
>         ColumnFixedString(16, values);
> ```

Hi, this maybe the minimum change. Or, windows dll will be impossible
```c++
template<typename Values>
    static ColumnFixedString MakeColumnFixedString(size_t n, const Values& values) {
        ColumnFixedString cfs(n);
        for (const auto& v : values) {
            cfs.Append(v);
        }
        return cfs;
    }
···

It will become this:
```c++
const auto values = std::array<const char*, 2>{
                "first row",
                "second row"};
        ColumnFixedString::MakeColumnFixedString(16, values);
```
@Enmk

1261385937: another attempt :
Add __declspec(dllexport) for each related class.  It is a terrible change.

Jihadist:
> @1261385937 Major question, why do you need to enable `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` ?

Windows shared library requires import library. Cmake can build it by default without specificing export symbols with `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`

1261385937:
> @1261385937 Major question, why do you need to enable `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` ?

Or should add __declspec(dllexport)  for almost whole class expicity

1261385937:
Windows static lib may be better.
I will close this draft.

PR #231, opened 2022-10-18
Implementation per query settings

den818:
I would like to suggest the implementation of per-query settings.  First of all, I refuse to support binary settings serialization because it is deprecated since server version 19.18.   The implementation is based on the following ideas 
1. Use compiled checked settings names and values (especially for enums). It will protect from mistyping.
2. Easy synchronization settings list with server source code.  That's why settings, menus, and defaults are just copy-pasted from server sources. All macros have the same syntax.
3. Settings have to be sent only if they have been set. No data is transmitted by default.  
4. Object QuerrySettings is light and easy to reuse in different queries. It contains unordered_map of string views of static names. So it allocate memory only for values.

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -645,12 +645,17 @@ void Client::Impl::SendQuery(const Query& query) {
         }
     }
 
-    /// Per query settings.
-    assert(server_info_.revision >=  DBMS_MIN_REVISION_WITH_SETTINGS_SERIALIZED_AS_STRINGS);
-    for(const auto& [name, field] : query.GetQuerySettings()) {
-        WireFormat::WriteString(*output_, name);
-        WireFormat::WriteVarint64(*output_, field.flags);
-        WireFormat::WriteString(*output_, field.value);
+    /// Per query settings
+    if (server_info_.revision >= DBMS_MIN_REVISION_WITH_SETTINGS_SERIALIZED_AS_STRINGS) {
+        for(const auto& [name, field] : query.GetQuerySettings()) {
+            WireFormat::WriteString(*output_, name);
+            WireFormat::WriteVarint64(*output_, field.flags);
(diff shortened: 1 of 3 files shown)


clickhouse/query.h
```diff
@@ -157,6 +145,7 @@ class Query : public QueryEvents {
 private:
     const std::string query_;
     const std::string query_id_;
+    std::shared_ptr<QuerySettings> query_settings_;
```
Enmk, line 148: why not just `QuerySettings` here and in the query methods, like:
```
    inline const QuerySettings& GetQuerySettings() const {
        return query_settings_;
    }

    /// Set per query settings
    inline Query& SetQuerySettings(QuerySettings query_settings) {
        query_settings_ = std::move(query_settings);
        return *this;
    }
```

clickhouse/client.cpp
```diff
@@ -644,14 +646,19 @@ void Client::Impl::SendQuery(const std::string& query, const std::string& query_
...
-    //else
+    if (query.GetQuerySettings()) {
+        assert(server_info_.revision >=  DBMS_MIN_REVISION_WITH_SETTINGS_SERIALIZED_AS_STRINGS);
+        for(const auto& [name, value] : *query.GetQuerySettings()) {
+            WireFormat::WriteString(*output_, name);
+            WireFormat::WriteVarint64(*output_, 0);
+            WireFormat::WriteString(*output_, value);
+        }
+    }
     WireFormat::WriteString(*output_, std::string());
```
Enmk, line 657:
```suggestion
    // Empty string signals end of serialized settings
    WireFormat::WriteString(*output_, std::string());
```

Enmk (changes requested):
Looks good except for some minor issues.

Please also consider adding tests with invalid setting key and/or values.

PR #232, opened 2022-10-21
support open telemetry

den818:


ut/client_ut.cpp
```diff
@@ -1097,6 +1098,52 @@ TEST_P(ClientCase, QuerySettings) {
...
+
+    Query query("INSERT INTO " + table_name + " (*) VALUES (\'Foo\'), (\'Bar\')" );
+    open_telemetry::TracingContext tracing_context;
+    std::srand(std::time(0));
+    tracing_context.trace_id.first  = std::rand();
+    tracing_context.trace_id.second = std::rand();
+    query.SetTracingContext(tracing_context);
+    client_->Execute(query);
+
+    try {
```
Enmk, line 1113: Sine this is not the first place were "SYSTEM FLUSH LOGS" is executed, it makes sense to factor out this code into a function/method, so there are no code duplicates.

Something like:
```
// client_ut.cpp
void FlushLogs(Client& client);

// later in the test:
FLusgLogs(*client_);
```

ut/client_ut.cpp
```diff
@@ -1097,6 +1098,52 @@ TEST_P(ClientCase, QuerySettings) {
...
 }
 
+TEST_P(ClientCase, TracingContext) {
+    Block block;
+    createTableWithOneColumn<ColumnString>(block);
+
+    Query query("INSERT INTO " + table_name + " (*) VALUES (\'Foo\'), (\'Bar\')" );
+    open_telemetry::TracingContext tracing_context;
+    std::srand(std::time(0));
+    tracing_context.trace_id.first  = std::rand();
```
Enmk, line 1108: Why not just `UUID(std::rand(), std::rand())` ?

clickhouse/base/string_utils.h
```diff
@@ -23,4 +26,37 @@ inline T FromString(const StringView& s) {
...
+template <>
+inline UUID FromString(const StringView& s) {
+   uint64_t f1 = 0, f2 = 0, f3 = 0, s1 = 0, s2 = 0;
+   std::scanf(s.data(),UUIDFormat(), f1, f2, f3, s1, s2);
+   return {(f1 << 16) + (f2 << 8) + f3, (s1 << 48) + s2};
+
+}
+
+inline std::string ToString(const UUID& v) {
+   std::array<char,37> buff;
```
Enmk, line 53: Maybe resize `std::string` to 37 items and build result in that space?

clickhouse/base/string_utils.h
```diff
@@ -23,4 +26,37 @@ inline T FromString(const StringView& s) {
...
+      return "%.8lx-%.4lx-%.4lx-%.4lx-%.12lx";
+   } else {
+      return "%.8llx-%.4llx-%.4llx-%.4llx-%.12llx";
+   }
+}
+
+template <>
+inline UUID FromString<UUID>(const std::string& s) {
+   uint64_t f1 = 0, f2 = 0, f3 = 0, s1 = 0, s2 = 0;
+   std::scanf(s.c_str(),UUIDFormat(), f1, f2, f3, s1, s2);
```
Enmk, line 40: 1. `sscanf` maybe?
2. you should check the return code of the `sscanf`, since it may indicate that not all fields were read.
3. both overloads should share the code, `std::string`-overload should call `StringView`-overload
4. Please add some minor unit-tests for these `FromString` to ensure that implementation is correct. (both positive and error cases)

clickhouse/base/string_utils.h
```diff
@@ -23,4 +26,37 @@ inline T FromString(const StringView& s) {
...
+   uint64_t f1 = 0, f2 = 0, f3 = 0, s1 = 0, s2 = 0;
+   std::scanf(s.data(),UUIDFormat(), f1, f2, f3, s1, s2);
+   return {(f1 << 16) + (f2 << 8) + f3, (s1 << 48) + s2};
+
+}
+
+inline std::string ToString(const UUID& v) {
+   std::array<char,37> buff;
+   //ffff ff ff ss ssssss
+   std::snprintf(buff.data(), buff.size(),
```
Enmk, line 55: You should check `std::snprintf` return code for errors.

clickhouse/base/string_utils.h
```diff
@@ -23,4 +26,37 @@ inline T FromString(const StringView& s) {
    return result;
 }
 
+inline const char* UUIDFormat() {
+   if constexpr (sizeof(long) == 8) {
+      return "%.8lx-%.4lx-%.4lx-%.4lx-%.12lx";
```
Enmk, line 31: Please take a look at ["Format constants for the std::fscanf family of functions"](https://en.cppreference.com/w/cpp/header/cinttypes)

Looks like you need to use `PRIu64`/`SCNu64`.

clickhouse/client.cpp
```diff
@@ -643,6 +645,28 @@ void Client::Impl::SendQuery(const Query& query) {
         if (server_info_.revision >= DBMS_MIN_REVISION_WITH_VERSION_PATCH) {
             WireFormat::WriteUInt64(*output_, info.client_version_patch);
         }
+
+        if (server_info_.revision >= DBMS_MIN_REVISION_WITH_OPENTELEMETRY) {
+            if (query.GetTracingContext()) {
```
Enmk, line 650:
```
if (const auto & tracing_context = query.GetTracingContext() ) {
```

fixes both ambiguous naming `trace`/`tracing` and getting that context twice.

Enmk (changes requested):
Ok, but please address highlighted issues.

clickhouse/base/string_utils.h
```diff
@@ -23,4 +26,37 @@ inline T FromString(const StringView& s) {
    return result;
 }
 
+inline const char* UUIDFormat() {
```
Enmk, line 29: Also, since `UUIDFormat()`, `ToString(UUID)`, `FromString(..., UUID)` are used only in unit-tests, it makes sense to put those into the files that are used exclusively by test binary, not to the library headers.

den818: I think, this functions are pretty useful for library users.  

If someone needs to pass UUID as query params, the function `ToString(UUID) ` will be extremely useful.

Enmk: There are approximately **0** settings that are UUID. Users are most likely to don't use UUIDs string representation or have their own string serialization/deserialzation for UUIDs.
So usability of those functions is questionable at best.

Please move those functions to `ut`.

Enmk (approved):
LGTM

PR #254, opened 2022-11-27
Support column type: map

den818:
This type described  here 
https://clickhouse.com/docs/en/sql-reference/data-types/map


1261385937:
👍

Enmk (approved):
LGTM

PR #257, opened 2022-11-30
Some improvement ColumnMap

den818:
That changes have two main purposes.

First of all to refuse of allocating memory for an intermediate container in method   `ColumnMap<K, V>:: Append(const T& value)`. A `ProjectedIterator` has been implemented for it. Implementation of this iterator is inspired by  [`boost::transform_iterator`](https://www.boost.org/doc/libs/master/libs/iterator/doc/transform_iterator.html)  and has to be replaced by `std::indirectly_readable` iterator.

The second purpose is to reduce memory allocation in `MapView::operator==(...)`. Vector integer indexes are used instead of copies of values.


Enmk (changes requested):
Could you please add some description and rationale behind the changes?

clickhouse/columns/map.h
```diff
@@ -187,13 +191,16 @@ class ColumnMapT : public ColumnMap {
...
-            auto comp = [](const auto& l, const auto& r) { return l.frist < r.first; };
-            std::sort(l.begin(), l.end(), comp);
-            std::sort(r.begin(), r.end(), comp);
-            return std::equal(l.begin(), l.end(), r.begin(), r.end());
+            const auto make_index = [](const auto& data) {
+                std::vector<size_t> result{data.Size()};
+                std::generate(result.begin(), result.end(), [i = 0] () mutable { return i++; });
+                std::sort(result.begin(), result.end(), [&data](size_t l, size_t r) {return data[l] < data[r];});
+                return result;
+            };
```
Enmk, line 199: Not entirely clear peace of code, so could you please add some comments to it?

Another thing: this allocates `(2 * N)*sizeof(size_t)` extra memory on the heap. IMO it could be more beneficial to create a sorted array (of pointers/references) from left items and then do a binary-search lookups with items from the right.

That way the algorithmic complexity stays the same `O(N*logN)`, but memory requirement drops significantly: only `N*sizeof(size_t)` instead of `2N*sizeof(t)`.

Anyway, it would be really nice to see some numbers on that, could you please add an extra entry to `clickhouse-cpp/ut/performance_tests.cpp` ?

clickhouse/columns/map.h
```diff
@@ -214,13 +221,17 @@ class ColumnMapT : public ColumnMap {
...
-        }
-        typed_data_->Append(container.begin(), container.end());
+        using BaseIter = typename T::const_iterator;
+        using KeyOfT = decltype(std::declval<BaseIter>()->first);
+        using ValOfT = decltype(std::declval<BaseIter>()->second);
+        using Functor = std::function<std::tuple<KeyOfT, ValOfT>(const BaseIter&)>;
+        using Iterator = ProjectedIterator<Functor, BaseIter>;
+
+        Functor functor = [](const BaseIter& i) {
+            return std::make_tuple(i->first, i->second);
```
Enmk, line 231: Just a random idea: maybe use `std::ref`/`std::cref` here? Because otherwise you are still copying values, not in bulk, but one-by-one.

```diff
-  return std::make_tuple(i->first, i->second);
+  return std::make_tuple(std::cref(i->first), std::cref(i->second));
```

Enmk (changes requested):
Not bad, but some changes are still required. Please see notes.

Enmk:
Closing to re-open and re-trigger CI/CD

Enmk (approved):
LGTM

Issue #153, opened 2022-02-11
Parameters in queries


AndreyAlifanov:
Does this client support parameterized queries?
I see that command line client can do this, but it uses it`s own implementation.

Enmk:
Hi Andey! Thank you for asking, right now clickhouse-cpp doesn't support parametrized queries. You are wellcome to submit any of (or all) a PR/API/implementation draft.

huotianyu:
> Hi Andey! Thank you for asking, right now clickhouse-cpp doesn't support parametrized queries. You are wellcome to submit any of (or all) a PR/API/implementation draft.

when support parameterized queries? 
when I excuted client.Select, I can not deal data.

Enmk:
Hi @huotianyu and sorry for a late reply. There are no solid date on this, seems like there is not enough interest from community.

To get things started, could you propose an API for such a feature?

gjergjk71:
Hi @Enmk, @AndreyAlifanov.

I was thinking maybe we could introduce parameters in queries in format: 

- `{<name>:<data type>}`
eg: `SELECT * FROM default.numbers WHERE id = {paramId:Int} AND name = {paramName:String}
`which would require either setting parameters on the sessions by executing:
`set "param_paramId" = 1; set "param_paramName" = 'one';`
and/or following the prepared-statement concept:
```
QueryStatement prep_stmt = client.PrepareStatement("SELECT * FROM default.numbers WHERE id = {paramId:Int} AND name = {paramName:String}");
prep_stmt->setInt("paramId", 1);
prep_stmt->setString("paramName", "one");
prep_stmt.execute([] (const Block& block)
    {
        for (size_t i = 0; i < block.GetRowCount(); ++i) {
            std::cout << block[0]->As<ColumnUInt64>()->At(i) << " "
                      << block[1]->As<ColumnString>()->At(i) << "\n";
        }
    }
);
```
`setInt`, `setString` - `QueryStatement` methods can either translate into raw queries or be replaced as-is on the string?

AndreyAlifanov:
Hi guys!

I think about it in terms of usual syntax for all DBMS, I know.
Something like this: `select * from my_table where id = ? and key > ?`.
Here **?** are parameters, that one can substitute with real values while preparing query.

Proposal of @gjergjk71 is good too, but I would prefer standard SQL-syntax.

gjergjk71:
[...]

PR #260, opened 2022-12-06
add parameterized syntax, just syntax-sugar

1261385937:
Now clickhouse-cpp supports parameterized syntax, and the placeholder should be ? .
It is just syntax-sugar, all replacing happens on the client side

Any document about clickhouse-server prepare statement? 
@Enmk

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -902,22 +902,6 @@ void Client::Execute(const Query& query) {
     impl_->ExecuteQuery(query);
 }
 
-void Client::Select(const std::string& query, SelectCallback cb) {
-    Execute(Query(query).OnData(std::move(cb)));
-}
-
-void Client::Select(const std::string& query, const std::string& query_id, SelectCallback cb) {
-    Execute(Query(query, query_id).OnData(std::move(cb)));
-}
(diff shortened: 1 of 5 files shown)


1261385937:
#153

Enmk:
Closing to re-open and re-trigger CI/CD

Enmk:
I suggest utilizing existing CH features for that:
https://clickhouse.com/docs/en/sql-reference/syntax#defining-and-using-query-parameters
For users that means they'd have to use `{pid:UInt16}` syntax.

For developers, that means we'd have to either
- prepend query with multiple `SET` queries, which I dislike.
- or update `TCPHandler` to take parameters from settings and apply that to query context (pretty much the same way it is done in HTTP handler)

Notes to myself on how to implement later on CH side:
src/Parsers/ASTSetQuery.h:10 `constexpr char QUERY_PARAMETER_NAME_PREFIX[] = "param_";`
Context::setQueryParameter() with those in TCPHandler.cpp
src/Interpreters/executeQuery.cpp `executeQuery` seems to be handling query parameters just fine.

1261385937:
ok

PR #262, opened 2022-12-09
Handle "Bool" columns

ays7:
Bool received columns are not recognized and mapped to "Void".
Mapping updated to treat received Bool columns as UInt8


Enmk (changes requested):
Thank you for contribution, please add some tests. 

At least some in `/ut/CreateColumnByType_ut.cpp`, and some that receives/sends data to CH server to Bool column (`ut/client_ut.cpp`).

ays7:
@Enmk sure
done

PR #268, opened 2022-12-14
Support get value at.

huyphams:
- Added get value at function, along with operator [].

Reason:

The current implement operator [] does not work in Objective-C, so adding the ->At().

Update:
Sorry, my bad, it works with the syntax:  

``` 
auto cols = tup->operator[](c); 
````
but it's a bit ugly IMO.


Enmk (approved):
LGTM

PR #270, opened 2022-12-18
fix bug in ColumnLowCardinality::Load

den818:
`ColumnLowCardinality::Load` allocate the wrong amount rows for `LowCardinality(Nullable(X))` columns. In most cases, it leads to extra memory allocation than required. But in case then the number of rows in a column is less than the dictionary size exception "`can't load column 'some column' of type LowCardinality(Nullable(X))`" happened. The dictionary always contains null and default values. So if the column contains every unique value once and doesn't contain a null or default value, then such a situation happens.

### DIFF
--- a/clickhouse/columns/lowcardinality.cpp
+++ b/clickhouse/columns/lowcardinality.cpp
@@ -281,7 +281,7 @@ auto Load(ColumnRef new_dictionary_column, InputStream& input, size_t rows) {
 
     if (auto nullable = new_dictionary_column->As<ColumnNullable>()) {
         nullable->Append(true);
-        for(std::size_t i = 1; i < new_index_column->Size(); i++) {
+        for(std::size_t i = 1; i < dataColumn->Size(); i++) {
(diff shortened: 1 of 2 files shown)


huyphams:
I'm still getting this issue, using the latest commit https://github.com/ClickHouse/clickhouse-cpp/commit/d03dbfe855d4fd34f52038cc688fc4c753f14492

Run this query:

```
SELECT * FROM "system"."session_log" LIMIT 300 OFFSET 0;
```

Error message:
can't load column 'roles' of type Array(LowCardinality(String))

Server version: ClickHouse 21.12.3.32

den818:
Hi [huyphams](https://github.com/huyphams)

It is another bug. It will be fixed in PR 

https://github.com/ClickHouse/clickhouse-cpp/pull/272

huyphams:
Thanks @den818

huyphams:
I confirmed it works @den818 thank you so much.

PR #278, opened 2023-01-05
Fix getaddrinfo error reporting

nUl1:
`getaddrinfo` error reporting is currently broken on non-Windows platforms as only a subset of its errors are reported via `errno`.
(On Windows everything gets combined in `WSAGetLastError()` and `gai_strerror` is declared not thread-safe for some reason.)

### DIFF
--- a/clickhouse/base/socket.h
+++ b/clickhouse/base/socket.h
@@ -60,6 +60,18 @@ public:
 
 #endif
 
+#if defined(_unix_)
+
+class getaddrinfoErrorCategory : public std::error_category {
+public:
+    char const* name() const noexcept override final;
+    std::string message(int c) const override final;
+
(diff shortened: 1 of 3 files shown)


clickhouse/base/socket.h
```diff
@@ -60,6 +60,18 @@ class windowsErrorCategory : public std::error_category {
 
 #endif
 
+#if defined(_unix_)
+
+class getaddrinfoErrorCategory : public std::error_category {
```
Enmk, line 65: Do you have to make it part of the API, i.e. visible to the users? IMO it would be better to move declaration to .cpp

nUl1: There's no requirement for it to be public. But users will have access to it through `e.code().category()` and nothing to compare it to.

PR #281, opened 2023-02-02
Add options to use system abseil, lz4 and cityhash

### LINKED ISSUE #86: relative import error

### LINKED ISSUE #99: Current setup with absl is dangerous

Jihadist:
> This change closes #86, and closes #99.
> 
> Furthermore, it eases Conan packaging, as Conan already provides abseil, lz4 and cityhash.
> 
> Signed-off-by: David Keller [david@woorton.com](mailto:david@woorton.com)

It's just rebased #168


Enmk (approved):
LGTM

PR #298, opened 2023-04-16
Dont run query id tests on non linux

Enmk:
Fixes red build on MacOs and Windows


Enmk:
closing to re-open and re-trigger builds against most recent master

PR #301, opened 2023-04-19
Fix system cityhash setup

ZhongRuoyu:
cityhash does not provide any CMake packages. Add a `Findcityhash.cmake` script to help find cityhash.

### DIFF
--- a/cmake/Findcityhash.cmake
+++ b/cmake/Findcityhash.cmake
@@ -1,0 +1,26 @@
+find_path(cityhash_INCLUDE_DIR
+  NAMES city.h
+  DOC "cityhash include directory")
+mark_as_advanced(cityhash_INCLUDE_DIR)
+find_library(cityhash_LIBRARY
+  NAMES cityhash libcityhash
+  DOC "cityhash library")
+mark_as_advanced(cityhash_LIBRARY)
+
+# Unlike lz4, cityhash's version information does not seem to be available.
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(cityhash
+  REQUIRED_VARS cityhash_LIBRARY cityhash_INCLUDE_DIR)
+
+if (cityhash_FOUND)
+  set(cityhash_INCLUDE_DIRS "${cityhash_INCLUDE_DIR}")
+  set(cityhash_LIBRARIES "${cityhash_LIBRARY}")
+
+  if (NOT TARGET cityhash::cityhash)
+    add_library(cityhash::cityhash UNKNOWN IMPORTED)
+    set_target_properties(cityhash::cityhash PROPERTIES
+      IMPORTED_LOCATION "${cityhash_LIBRARY}"
+      INTERFACE_INCLUDE_DIRECTORIES "${cityhash_INCLUDE_DIR}")
+  endif ()
+endif ()


ZhongRuoyu:
Homebrew does provide the `cityhash` package :) . We are already working around this in https://github.com/Homebrew/homebrew-core/pull/128575, with the following adjustments:

```ruby
    # `cityhash` does not provide a pkg-config or CMake config file.
    # Help CMake find it.
    inreplace "CMakeLists.txt", "FIND_PACKAGE(cityhash REQUIRED)",
                                "FIND_LIBRARY(CITYHASH NAMES cityhash REQUIRED)"
    inreplace "clickhouse/CMakeLists.txt", "cityhash::cityhash", "cityhash"
```

With this, the build passed on both macOS and Linux.

I suggest trying the following in a `homebrew/ubuntu22.04` Docker container, to verify the fix:

```
brew install cityhash cmake

git clone https://github.com/ClickHouse/clickhouse-cpp.git
cd clickhouse-cpp

# This should fail
cmake -S . -B build -DWITH_SYSTEM_CITYHASH=ON -DCMAKE_FIND_ROOT_PATH="$(brew --prefix)"

# This should work
git fetch origin refs/pull/301/head:fix
git checkout fix
export CXXFLAGS="-I`brew --prefix cityhash`/include"
cmake -S . -B build -DWITH_SYSTEM_CITYHASH=ON -DCMAKE_FIND_ROOT_PATH="$(brew --prefix)"
cmake --build build -j "$(nproc)"
```

Enmk:
Nice! Could you please add a workflow that verifies such setup? That would be a more solid gurantee that it doesn't break in the future :)

ZhongRuoyu:
Sure, I've pushed 45edfc1 and fefc0ee. I also adjusted the patch to avoid additional `-L` linker flags, because it seems that `/usr/local/lib` is not in macOS's default linker search paths ([error](https://github.com/ZhongRuoyu/clickhouse-cpp/actions/runs/4767403461/jobs/8475586361#step:5:51) here).

However, when testing this on my fork, some of the tests failed. Specifically, [these on Linux](https://github.com/ZhongRuoyu/clickhouse-cpp/actions/runs/4772319533/jobs/8484756130#step:10:1270), and [these on macOS](https://github.com/ZhongRuoyu/clickhouse-cpp/actions/runs/4772319532/jobs/8484755804#step:7:1253). I suspect that this is due to a mismatch between the cityhash versions.

Enmk:
> According to cityhash's [changelog](https://github.com/google/cityhash/blob/master/NEWS), cityhash v1.1 did "Change existing functions to improve their hash quality and/or speed". Does this raise a concern?

Breaking a hash compatibility with CH is a big "no-no", perhaps you can choose another version from homebrew to install?

Enmk:
https://github.com/ClickHouse/clickhouse-cpp/actions/runs/4772342520/jobs/8517401001?pr=301 - `RemoteTLS/ReadonlyClientTest.Select/0` seems unrelated
https://github.com/ClickHouse/clickhouse-cpp/actions/runs/4772342534/jobs/8517401004?pr=301
ArrayOfLowCardinality.InsertAndQuery
GenericColumnTest/0.RoundTrip
GenericColumnTest/0.NulableT_RoundTrip
GenericColumnTest/1.RoundTrip
GenericColumnTest/1.NulableT_RoundTrip
`closed: The operation completed successfully.`

CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ PROJECT (CLICKHOUSE-CLIENT)
     ENDIF ()
 
     IF (WITH_SYSTEM_CITYHASH)
-        FIND_PACKAGE(cityhash REQUIRED)
```
Jihadist, line 80: Please do not use find_library, you'll break compatibility with package managers. Look at https://github.com/ClickHouse/clickhouse-cpp/pull/168#issuecomment-1367840372

ZhongRuoyu: @Jihadist could you please elaborate on "compatibility with package managers"? cityhash does not ship any CMake config file (not even a pkg-config one), so it doesn't seem impossible to find `cityhash` with `find_package`.

Disclaimer: I am a maintainer of the [Homebrew](https://brew.sh/) package manager. In https://github.com/Homebrew/homebrew-core/pull/128575, we could not find `cityhash` without fixing this. (Though now we are going to use the vendored `cityhash` due to hash compatibility concerns as mentioned in https://github.com/ClickHouse/clickhouse-cpp/pull/301#issuecomment-1520592157. But that's a different point.)

ZhongRuoyu:
OK, I think I'll need to roll back the workflow changes because I don't know any system that ships this very version of cityhash (v1.0.2). (The fact that the clickhouse library was built should suggest that the changes are fine, though.)

As for the other failures, I am very confused because I haven't touched the Windows / MinGW workflows at all. Anyway, I've tried to keep only the CMake changes in [another testing branch](https://github.com/ZhongRuoyu/clickhouse-cpp/tree/system-cityhash-workflow); the tests should work fine.

CMakeLists.txt
Jihadist: @ZhongRuoyu  vcpkg and conan provide cityhash as _cityhash_ cmake target. I mentioned comment especially because author provide an example of conanfile for clickhouse-cpp (it was not actually merged but it another question)

ZhongRuoyu: Well, I'd argue that it's not the general case. As I mentioned earlier, cityhash does *not* provide any config files, so it is the user's (project's) responsibility to find out where the cityhash library is. To be clearer: package managers can provide their own CMake target, but I don't think projects should rely on that, because it is not the general case. For instance, none of Homebrew, [FreeBSD](https://github.com/freebsd/freebsd-ports/blob/main/devel/cityhash/pkg-plist), or [Raspbian](http://archive.raspbian.org/raspbian/pool/main/c/cityhash/) provide anything more than the headers and the library.

Nevertheless, since it wouldn't hurt anything, let me unbreak that by attempting `find_package` first.

clickhouse/CMakeLists.txt
```diff
@@ -38,12 +38,18 @@ IF (WITH_OPENSSL)
     LIST(APPEND clickhouse-cpp-lib-src base/sslsocket.cpp)
 ENDIF ()
 
+IF (WITH_SYSTEM_CITYHASH AND NOT cityhash_FOUND)
+    SET (CITYHASH ${SYSTEM_CITYHASH})
```
Enmk, line 42: Perhaps we can add something similar to https://github.com/ClickHouse/clickhouse-cpp/blob/master/cmake/Findlz4.cmake to properly set a `cityhash::cityhash` imported library instead?

CMakeLists.txt
Jihadist: > As I mentioned earlier, cityhash does not provide any config files, so it is the user's (project's) responsibility to find out where the cityhash library is. To be clearer: package managers can provide their own CMake target, but I don't think projects should rely on that, because it is not the general case. For instance, none of Homebrew, [FreeBSD](https://github.com/freebsd/freebsd-ports/blob/main/devel/cityhash/pkg-plist), or [Raspbian](http://archive.raspbian.org/raspbian/pool/main/c/cityhash/) provide anything more than the headers and the library.

Yep, you are right here. But it's not an argument to use `find_library `in root cmakelists. `find_library `should be used inside custom cmake find*.cmake script and this script may import target. So you do not need to remove find_package, and it will work with your homebrew or bsd package.

Jihadist: Your current solution is not bad too. Maybe it ok

clickhouse/CMakeLists.txt
ZhongRuoyu: That's actually a good idea. I pushed a `Findcityhash.cmake` and that should do it. I've verified this with a workflow in my fork: https://github.com/ZhongRuoyu/clickhouse-cpp/actions/runs/4840656304/jobs/8626442071#step:7:41. (Failure is due to a mismatched cityhash version, as discussed earlier -- but it's functional.)

CMakeLists.txt
ZhongRuoyu: Thanks -- @Enmk also pointed this out. This was out of my knowledge, but indeed it looks more compatible with the current CMake configs. I've pushed a `Findcityhash.cmake` script and that should help to find cityhash.

Jihadist (approved):
Looks nice, thank you!

Enmk:
Temporary closing to re-trigger against latest master

ZhongRuoyu:
Thanks for the help to get this merged, @Enmk @Jihadist!

PR #310, opened 2023-05-25
Multiple endpoints for connection.

MikhailBurdukov:
Specify multiple hosts for connection, to provide fault tolerance. 
```
.SetHosts({ getEnvOrDefault("CLICKHOUSE_HOST",      "asasdasd"), 
                            getEnvOrDefault("CLICKHOUSE_HOST",      "localhost"),
                            getEnvOrDefault("CLICKHOUSE_HOST",      "noalocalhost"), 
                            getEnvOrDefault("CLICKHOUSE_HOST",      "localhost"), 
                           })
```
The endpoint is selecting according to `EndpointsIterationAlgorithm`, rn only RoundRobin approach is implemented.

Ref #139


clickhouse/base/socket.cpp
```diff
@@ -390,9 +390,9 @@ std::unique_ptr<OutputStream> Socket::makeOutputStream() const {
 
 NonSecureSocketFactory::~NonSecureSocketFactory()  {}
 
-std::unique_ptr<SocketBase> NonSecureSocketFactory::connect(const ClientOptions &opts) {
-    const auto address = NetworkAddress(opts.host, std::to_string(opts.port));
+std::unique_ptr<SocketBase> NonSecureSocketFactory::connect(const ClientOptions &opts, const std::shared_ptr<EndpointsIteratorBase> endpoints_iterator) {
```
aalexfvk, line 393: What do you think about passing `NetworkAddress` object to this function instead of shared pointer to iterator ? Iterator seems redundant here

clickhouse/client.cpp
```diff
@@ -194,30 +207,55 @@ class Client::Impl {
     std::unique_ptr<InputStream> input_;
     std::unique_ptr<OutputStream> output_;
     std::unique_ptr<SocketBase> socket_;
+    std::shared_ptr<EndpointsIteratorBase> endpoints_iterator;
```
aalexfvk, line 210: Do we really need to store shared pointer to iterator here ? Or we can create it as needed. It looks pretty lightweight

MikhailBurdukov: For the Round Robin algo, we have to keep the same instance of the object, because otherwise `current_index` will be reset to zero every time the object is created.

clickhouse/base/socket.cpp
MikhailBurdukov: Agree, fixed

clickhouse/client.h
```diff
@@ -58,6 +67,14 @@ struct ClientOptions {
     /// Service port.
     DECLARE_FIELD(port, unsigned int, SetPort, 9000);
 
+    /// Hostnames of the servers. The next host to connect is selected according to the EndpointsIterationAlgorithm.
+    /// Note: If SetHost and SetHosts are setted, host will be placed at the beginning of the hosts vector.  
+    DECLARE_FIELD(hosts, std::vector<std::string>,  SetHosts, std::vector<std::string>());
+    /// Ports of the servers.
+    DECLARE_FIELD(ports, std::vector<unsigned int>, SetPorts, std::vector<unsigned int>());
+    /// Algorithm for selecting the next endpoint for connection.
+    DECLARE_FIELD(iteration_algo, EndpointsIterationAlgorithm, SetEndpointsIterationAlgorithm, EndpointsIterationAlgorithm::RoundRobin);
```
Enmk, line 76: We don't need to expose this setting at this point, since there is no meaningful way changing it to something else. It only makes sense to expose those when we have multiple algorithms.

ut/client_ut.cpp
```diff
@@ -1210,3 +1210,87 @@ INSTANTIATE_TEST_SUITE_P(ClientLocalFailed, ConnectionFailedClientTest,
...
+        FAIL() << "Got an unexpected exception : " << e.what();
+    }
+}
+
+
+INSTANTIATE_TEST_SUITE_P(ClientMultipleEndpoints, ConnectionSuccessTestCase,
+    ::testing::Values(ClientCase::ParamType{
+        ClientOptions()
+            .SetHosts({
+                    getEnvOrDefault("CLICKHOUSE_HOST",     "somedeadhost"),
```
Enmk, line 1234: Please use  plain `"somedeadhost"` value instead (without `getEnvOrDefault`), so it remains invalid host name on CI/CD.

clickhouse/base/endpoints_iterator.cpp
```diff
@@ -0,0 +1,43 @@
+#include "endpoints_iterator.h"
+#include "../client.h"
+
+namespace clickhouse {
+
+RoundRobinEndpointsIterator::RoundRobinEndpointsIterator(const ClientOptions& opts) :
+     hosts (opts.hosts)
```
Enmk, line 7: IMO you should put value from 'opts.host' here as first value, otherwise it is not used at all.

clickhouse/client.cpp
myrrc: I agree with Vasily and believe endpoints iteration-related code could be simplified. Having a base class overridden by a specific algorithm is a good general solution, but in case only one algorithm is implemented, I think it's better to have a dead simple `RoundRobinEndpointsIterator endpoints_iterator` without inheritance and smart pointers.

If you still want a general solution, consider using `std::unique_ptr` as we seem to have only one user of this field.

myrrc (changes requested):

MikhailBurdukov:
In the last commit, a behavior was added that allows users to reset the connection to a connected endpoint, which is described [here](https://github.com/ClickHouse/clickhouse-cpp/pull/139/files#r790480807).  I also checked manually that the client is trying to reconnect to the same endpoint using `options_.send_retries`, but I don't know how to implement this in the test, because as far as I understand, there are no tests that allow you to manipulate the connection or the `ClickHouse-server' directly. Does anyone have some ideas how to test such situation?

clickhouse/client.h
```diff
@@ -44,6 +44,19 @@ enum class CompressionMethod {
     LZ4     =  1,
 };
 
+struct Endpoint {
+    std::string host;
+    unsigned int port = 9000;
```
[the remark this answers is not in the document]
Enmk: Yep, I think this is a good idea, even though this alters public interface

clickhouse/client.cpp
```diff
@@ -64,8 +64,12 @@ struct ClientInfo {
 };
 
 std::ostream& operator<<(std::ostream& os, const ClientOptions& opt) {
-    os << "Client(" << opt.user << '@' << opt.host << ":" << opt.port
-       << " ping_before_query:" << opt.ping_before_query
+    os << "Client(";
+    for (size_t i = 0; i < opt.endpoints.size(); i++)
```
Enmk, line 68: that ignores `opt.host` and `opt.port` completely

clickhouse/base/socket.h
```diff
@@ -88,7 +89,7 @@ class SocketFactory {
 
     // TODO: move connection-related options to ConnectionOptions structure.
 
-    virtual std::unique_ptr<SocketBase> connect(const ClientOptions& opts) = 0;
+    virtual std::unique_ptr<SocketBase> connect(const ClientOptions& opts, const std::string& host, const std::string& port) = 0;
```
Enmk, line 92:
```suggestion
    std::unique_ptr<SocketBase> connect(const ClientOptions& opts, const Endpoint & endpoint) = 0;
```

clickhouse/base/endpoints_iterator.h
```diff
@@ -0,0 +1,48 @@
...
+   virtual void Next() = 0;
+   // Get the address of current endpoint. 
+   virtual std::string GetHostAddr() const = 0;
+
+   // Get the port of current endpoint.
+   virtual uint16_t GetPort() const = 0;
+
+   // Reset iterations.
+   virtual void ResetIterations() = 0; 
+   virtual bool NextIsExist() const = 0;
```
Enmk, line 27:
```suggestion
   virtual std::optional<Endpoint> Next() = 0;    // nullopt if no more endpoints
   virtual void ResetIterations() = 0;    // start over
```

Cleaner and much simpler API

clickhouse/client.cpp
```diff
@@ -194,32 +210,36 @@ class Client::Impl {
...
 
     ServerInfo server_info_;
 };
 
+ClientOptions modifyClientOptions(ClientOptions opts)
+{
+    if (opts.host.empty())
+        return opts;
+
+    Endpoint endpoint_single({opts.host, opts.port});
```
Enmk, line 225:
```suggestion
    Endpoint default_endpoint({opts.host, opts.port});
```

clickhouse/client.cpp
```diff
@@ -194,32 +210,36 @@ class Client::Impl {
...
+    if (std::find(opts.endpoints.begin(), opts.endpoints.end(), endpoint_single) == std::end(opts.endpoints)) {
```
Enmk, line 226: IMO it is best to make `default_endpoint` first endpoint unconditionally, so it is simple, easy and predictable

clickhouse/client.cpp
```diff
@@ -64,8 +64,12 @@ struct ClientInfo {
 };
 
 std::ostream& operator<<(std::ostream& os, const ClientOptions& opt) {
-    os << "Client(" << opt.user << '@' << opt.host << ":" << opt.port
-       << " ping_before_query:" << opt.ping_before_query
+    os << "Client(";
```
Enmk, line 67: That ignores `opt.host` and `opt.port` completely, and this `operator<<` might be used on unmodified `ClientOptions` (by tests or users), making output confusing.

clickhouse/base/endpoints_iterator.h
MikhailBurdukov: Yes, but `ResetIterations` is needed in order not to create an infinite loop for endpoint choosing.
Otherwise we have to  check it thought `opts.endpoints.size()`,which doesn't look like a good solution. Any ideas how to do so?

clickhouse/client.cpp
```diff
@@ -111,6 +115,10 @@ std::unique_ptr<SocketFactory> GetSocketFactory(const ClientOptions& opts) {
         return std::make_unique<NonSecureSocketFactory>();
 }
 
+std::unique_ptr<EndpointsIteratorBase> GetEndpointsIterator(const std::vector<Endpoint>& endpoints) {
```
Enmk, line 118:
```suggestion
std::unique_ptr<EndpointsIteratorBase> GetEndpointsIterator(const ClientOptions& options) {
```

MikhailBurdukov:
Discussed in direct messages:
- [x] Connection format: N times to all hosts. Example with 3 endpoints and N = 2 : 1- 2 -3 -1 - 2 -3.
- [x] Iterator must have only one method: `virtual Endpoint Next() = 0;` and it is an endless generator of endpoints.
- [x] The logic of connection control in `Client Impl`. Number of connection attempts = endpoints.size() * send_retry
- [x] If the host has become unavailable, we first try to recreate the connection with it, if it did not work, then we try to find a new one.

All this is done in the last commit

PR #311, opened 2023-06-10
fix compile warning about c++17

1261385937:
For win platform, fix compile warning about c++17.
Too many warnings make me crazy


1261385937:
make a new pr

PR #328, opened 2023-09-14
Add move semantics for ColumnEnum

bduminiuc:
The `ColumnEnum<T>` can be initialized with `std::vector<T>` by moving the vector instead of copying.


Enmk (approved):
LGTM, test failures are unrelated.

`Client/ClientCase.TracingContext` test will be fixed in a separate PR

PR #330, opened 2023-09-19
Add reserve, capacity methods to numeric columns

rtcote:
- Reserve & Capacity are to enable preallocating large data sets
- SwapElements is to enable in-place sorting of elements in columns
- CompareElementsLT &  CompareElementsGT to determine sort order efficiently

### DIFF
--- a/clickhouse/columns/column.h
+++ b/clickhouse/columns/column.h
@@ -80,12 +80,6 @@ public:
     /// Returns count of rows in the column.
     virtual size_t Size() const = 0;
 
-    /// Increase the capacity of the column (currently only supported by numeric/date types at this time)
-    virtual void Reserve([[maybe_unused]]size_t new_cap) {}
-
-    /// Returns the capacity of the column (currently only supported by numeric/date types at this time)
-    virtual size_t Capacity() const { return Size(); }
-
     /// Makes slice of the current column.
     virtual ColumnRef Slice(size_t begin, size_t len) const = 0;
 
--- a/clickhouse/columns/column.h
+++ b/clickhouse/columns/column.h
@@ -80,10 +80,10 @@ public:
     /// Returns count of rows in the column.
     virtual size_t Size() const = 0;
 
-    /// Increase the capacity of the column (currently only supported by numeric types at this time)
+    /// Increase the capacity of the column (currently only supported by numeric/date types at this time)
     virtual void Reserve([[maybe_unused]]size_t new_cap) {}
 
-    /// Returns the capacity of the column (currently only supported by numeric types at this time)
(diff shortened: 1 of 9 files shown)


Enmk (changes requested):
Thank you for contributing, there are few minor things:
- Please add tests to verify correctness of new methods
- I doubt that we should add 'Reserve' and 'Capacity' as virtual methods to Column. It would remain unimplemented for some of the column types (like map, array, and low cardinality) for some substantial amount of time.
There are few options:
- Move those to the ColumnVector
- Add method 'std::vector<T> & ColumnVector::Data()' to allow direct manipulation
- pre-populate vector before moving it into column.

I personally prefer the third one, but that is just me.

rtcote:
Test cases added for Reserve & Capacity and SwapElements methods

rtcote:
@Enmk what else should I do at this point regarding this PR?  Thank you

Enmk (changes requested):
I still very dislike the proposed interfaces, they are incoherent and unnecessarily complex.

You should know the type of the column you are inserting into, so making `Reserve()` and `Capacity()` part of the `Column` interface is overkill -- some of the columns would never be able to conform to that interface. And in the places where you know that your column supports it, you'll downcast it to a proper type before doing multiple `Append()`s anyway.

Second, why do you need the sorting methods? If you are planning to sort the entire block according to some key column, then perhaps it is better to do done on the server side.

If you want to sort only a single column, perhaps it is better to do over raw vector rather than via `SwapElements`, `CompareElementsLT`, `CompareElementsGT`. Moreover, those comparisons are not adjustable enough. E.g. what if you would need to have a case-insensitive comparison of strings for `ColumnString`?

IMO if you need a generic "sorting" or comparison, you can do that using `ItemView`s or direct access methods.

In conclusion: it is ok to add `direct access` methods, that act in the spirit of ClickHouse: don't do unnecessary work, and allow low-level access for doing things on concrete data.

I.e. obvious candidates: `std::vector<T> & ColumnVector::ModifyableData();`

rtcote:
@Enmk the sorting methods were only provided for numeric types.  Why am I sorting outside of the database, for efficiency of inserts into the database as I am dealing with large block inserts and I do not want the database to perform the sorting upon insert.

I avoided exposing the raw vector as that breaks the abstraction as it requires knowing what the implementation uses for the storage datatype of the ColumnVector in the client app code.  As an example, knowing that datetime is a uint32 and should that ever change then it breaks the client app code.  The original changes made did not break the abstraction.

My latest changes are inline with your review now.

Enmk (approved):
LGTM, some minor issues will be fixed after PR is merged.

rtcote:
Thank you @Enmk.  Is there anything else that I need to do before we merge it in?

PR #332, opened 2023-09-22
Add rvalue vector ctors for ColumnDate, ColumnDate32 and ColumnIPv4

bduminiuc:
These constructors can be valuable for enhancing performance in scenarios where a user knows the row count and can preallocate memory.

I also considered adding a similar constructor for `ColumnDateTime`, but its public interface exclusively utilizes `time_t`, while `uint32_t` is an implementation detail. I'm unsure if extending the interface with implementation details would be a good idea.

### DIFF
--- a/ut/columns_ut.cpp
+++ b/ut/columns_ut.cpp
@@ -199,6 +199,17 @@ TEST(ColumnsCase, Date32_Int32_interface) {
     ASSERT_EQ(col1->RawAt(2), -1234);
 }
 
+TEST(ColumnsCase, DateTime_construct_from_rvalue_data) {
+    auto const expected = MakeNumbers<uint32_t>();
+
+    auto data = expected;
+    auto col1 = std::make_shared<ColumnDateTime>(std::move(data));
+
+    ASSERT_EQ(col1->Size(), expected.size());
+    for (size_t i = 0; i < expected.size(); ++i) {
+        ASSERT_EQ(col1->At(i), static_cast<std::time_t>(expected[i]));
+    }
+}
 
 TEST(ColumnsCase, DateTime64_0) {
     auto column = std::make_shared<ColumnDateTime64>(0ul);
--- a/ut/columns_ut.cpp
+++ b/ut/columns_ut.cpp
@@ -228,10 +228,7 @@ TEST(ColumnsCase, DateTime_construct_from_rvalue_data) {
     auto data = expected;
     auto col1 = std::make_shared<ColumnDateTime>(std::move(data));
 
-    ASSERT_EQ(col1->Size(), expected.size());
-    for (size_t i = 0; i < expected.size(); ++i) {
-        ASSERT_EQ(col1->At(i), static_cast<std::time_t>(expected[i]));
-    }
+    EXPECT_TRUE(CompareRecursive(*col1, expected));
 }
 
 TEST(ColumnsCase, DateTime64_0) {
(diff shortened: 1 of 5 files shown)


Enmk (changes requested):
Hi @bduminiuc, thanks for your contibution!

As for `ColumnDateTime` -- feel free to add constructor that takes `vector<uint32_t> &&`, the underlying storage type is not going to be changed in the foreseeable future.

Also, please consider adding unit-tests that verify the added constructors.

bduminiuc:
Hi, @Enmk,

The constructor for `ColumnDateTime` and the unit tests mentioned earlier were added.

ut/columns_ut.cpp
```diff
@@ -185,6 +185,17 @@ TEST(ColumnsCase, Date_UInt16_interface) {
     ASSERT_EQ(col1->RawAt(1), 1234u);
 }
 
+TEST(ColumnsCase, Date_UInt16_construct_from_rvalue_data) {
+    auto const expected = MakeNumbers<uint16_t>();
+
+    auto data = expected;
+    auto col1 = std::make_shared<ColumnDate>(std::move(data));
+
+    ASSERT_EQ(col1->Size(), expected.size());
```
Enmk, line 194: you can use a single check instead of  for-loop here:
```C++
EXPECT_TRUE(CompareRecursive(*col1, expected));
```
(same for other places)

ut/columns_ut.cpp
```diff
@@ -199,6 +210,29 @@ TEST(ColumnsCase, Date32_Int32_interface) {
...
+    auto const expected = MakeNumbers<uint32_t>();
+
+    auto data = expected;
+    auto col1 = std::make_shared<ColumnDateTime>(std::move(data));
+
+    ASSERT_EQ(col1->Size(), expected.size());
+    for (size_t i = 0; i < expected.size(); ++i) {
+        ASSERT_EQ(col1->At(i), static_cast<std::time_t>(expected[i]));
+    }
+}
```
Enmk, line 235: you can use a single check instead of  for-loop here:
```C++
EXPECT_TRUE(CompareRecursive(*col1, expected));
```
Takes care of size checks, value checks, and forms a pretty explanation message if comparison fails.

ut/columns_ut.cpp
```diff
@@ -185,6 +185,17 @@ TEST(ColumnsCase, Date_UInt16_interface) {
...
 
+TEST(ColumnsCase, Date_UInt16_construct_from_rvalue_data) {
+    auto const expected = MakeNumbers<uint16_t>();
+
+    auto data = expected;
+    auto col1 = std::make_shared<ColumnDate>(std::move(data));
+
+    ASSERT_EQ(col1->Size(), expected.size());
+    for (size_t i = 0; i < expected.size(); ++i) {
+        ASSERT_EQ(col1->RawAt(i), expected[i]);
```
Enmk, line 196: `CompareRecursive` isn't going to work due to value transformation by `ColumnDate`, same for `DateTime32`. methods returning bogus `time_t` values is the hard legacy we need to get rid of, but I haven't decided how to make a smooth transition for the users...

Enmk (commented):
Please address minor things in the unit tests and it is going to be good to merge.

ut/columns_ut.cpp
```diff
@@ -528,6 +562,23 @@ TEST(ColumnsCase, ColumnIPv4_construct_from_data)
     EXPECT_ANY_THROW(ColumnIPv4(ColumnRef(std::make_shared<ColumnString>())));
 }
 
+TEST(ColumnsCase, ColumnIPv4_construct_from_rvalue_data) {
+    std::vector<uint32_t> data = {
```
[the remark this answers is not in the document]
bduminiuc: The newly added constructor for `ColumnIPv4` accepts a vector of raw values `std::vector<uint32_t>`, while `MakeIPv4` returns `in_addr`. It seems that the `MakeIPv4` function cannot be used to provide test coverage for `ColumnIPv4(std::vector<uint32_t>&&)`.

Enmk (approved):
LGTM, will be merged after #336

Issue #335, opened 2023-09-27
Segfault caused by outputstream being null.


rtcote:
It appears that the output stream is getting set to a nullptr.  Was running v2.5.0 of the client.  When I rolled-back to v2.4.0 of the client, the issue is no longer present.  This issue shows up the first time I call Client::Insert() for each client instance.  It's not clear what I am doing on the application side to cause this.  Any thoughts as to how I can debug this?

My client is configured in the following manner:
```
m_client = std::make_shared<clickhouse::Client>(
    clickhouse::ClientOptions().SetHost(host)                                                                           // a hostname
                               .SetPingBeforeQuery(client_config.ping_before_query)                                     // true
                               .SetSendRetries(client_config.send_retries)                                              // 0
                               .SetCompressionMethod(compression_method)                                                // None
                               .SetRetryTimeout(std::chrono::seconds(client_config.retry_timeout_sec))                  // 0
                               .TcpKeepAlive(client_config.tcp_keep_alive)                                              // true
                               .SetConnectionConnectTimeout(std::chrono::seconds(client_config.connect_timeout_sec))    // 5
                               .SetConnectionRecvTimeout(std::chrono::milliseconds(client_config.recv_timeout_ms))      // 1000000
                               .SetConnectionSendTimeout(std::chrono::milliseconds(client_config.send_timeout_ms))      //  60000
                               );
```

Stack trace and gdb debugging
```
Thread 5 "rte-worker-8" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffc5c87000 (LWP 136)]
clickhouse::WireFormat::WriteAll (output=..., buf=<optimized out>, len=1) at /git/submodules/clickhouse-cpp/clickhouse/base/wire_format.cpp:36
36      /git/submodules/clickhouse-cpp/clickhouse/base/wire_format.cpp: No such file or directory.
(gdb) bt
#0  clickhouse::WireFormat::WriteAll (output=..., buf=<optimized out>, len=1) at /git/submodules/clickhouse-cpp/clickhouse/base/wire_format.cpp:36
#1  0x0000555559dce5c1 in clickhouse::WireFormat::WriteVarint64 (output=..., value=<optimized out>, value@entry=1) at /git/submodules/clickhouse-cpp/clickhouse/base/wire_format.cpp:86
[...]

PR #336, opened 2023-09-29
Fixed issue #335 by making at least one connection attempt

Enmk:
Due to setting `ClientOptions.SetSendRetries(0)`, Client made no attempts to connect to a server, which resulted in uninitialized streams and hence the crash.

Closes: #335

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -385,14 +385,15 @@ void Client::Impl::ResetConnectionEndpoint() {
 }
 
 void Client::Impl::CreateConnection() {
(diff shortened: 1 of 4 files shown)


rtcote:
@Enmk do you a rough idea of when this will be put into a versioned release?  Thank you!

Enmk:
> @Enmk do you a rough idea of when this will be put into a versioned release? Thank you!

it is already in [v2.5.1](https://github.com/ClickHouse/clickhouse-cpp/releases/tag/v2.5.1), released on 30th of September.

PR #340, opened 2023-10-26
optimize

1261385937:
Too many virtual function now :smile:. So add keyword 'final' for possible "de-virtualization" compiler optimize.  Most of the time, it is effective.


1261385937:
```cpp
class A {
public:
    virtual void f() = 0;
};

class B final: public A {
public:
    void f() override;
};

class C : public A {
public:
    void f() override;
};

void foo(B& b) {
    b.f();
}

void bar(C& c) {
    c.f();
}
```
and the simple asm 
```asm
bar(C&):
        mov     rax, QWORD PTR [rdi]
        jmp     [QWORD PTR [rax]]
foo(B&):
        jmp     B::f()
```

Enmk:
Hi @1261385937, thank you for the contribution!

Do you see any difference in performance and/or assembly after applying this patch?

1261385937:
@Enmk 

After days and nights of testing, the results have shown no significant improvement.
~~  :stuck_out_tongue_closed_eyes:

PR #341, opened 2023-10-27
add Reserve for column. Optimize large block insertion

1261385937:
For large block insertion(100,000+), preassign column memory

### DIFF
--- a/clickhouse/columns/column.h
+++ b/clickhouse/columns/column.h
@@ -52,6 +52,9 @@ public:
     /// Appends content of given column to the end of current one.
     virtual void Append(ColumnRef column) = 0;
 
+    /// Increase the capacity of the column for large block insertion.
+    virtual void Reserve(size_t new_cap) = 0;
+
     /// Template method to load column data from input stream. It'll call LoadPrefix and LoadBody.
     /// Should be called only once from the client. Derived classes should not call it.
     bool Load(InputStream* input, size_t rows);
--- a/clickhouse/columns/lowcardinality.cpp
+++ b/clickhouse/columns/lowcardinality.cpp
@@ -174,8 +174,7 @@ ColumnLowCardinality::ColumnLowCardinality(std::shared_ptr<ColumnNullable> dicti
 ColumnLowCardinality::~ColumnLowCardinality()
 {}
 
-void ColumnLowCardinality::Reserve(size_t new_cap)
-{
+void ColumnLowCardinality::Reserve(size_t new_cap) {
     dictionary_column_->Reserve(new_cap);
     index_column_->Reserve(new_cap);
 }
--- a/clickhouse/columns/lowcardinality.cpp
+++ b/clickhouse/columns/lowcardinality.cpp
@@ -174,6 +174,12 @@ ColumnLowCardinality::ColumnLowCardinality(std::shared_ptr<ColumnNullable> dicti
 ColumnLowCardinality::~ColumnLowCardinality()
 {}
 
+void ColumnLowCardinality::Reserve(size_t new_cap)
+{
+    dictionary_column_->Reserve(new_cap);
+    index_column_->Reserve(new_cap);
+}
+
 void ColumnLowCardinality::Setup(ColumnRef dictionary_column) {
     AppendDefaultItem();
 
--- a/clickhouse/columns/string.cpp
+++ b/clickhouse/columns/string.cpp
@@ -190,6 +190,13 @@ ColumnString::ColumnString(std::vector<std::string>&& data)
 ColumnString::~ColumnString()
 {}
 
+void ColumnString::Reserve(size_t new_cap)
+{
+    items_.reserve(new_cap);
+    // 16 is arbitrary number, assumption that string values are about ~256 bytes long.
+    blocks_.reserve(std::max<size_t>(1, new_cap / 16));
+}
+
 void ColumnString::Append(std::string_view str) {
     if (blocks_.size() == 0 || blocks_.back().GetAvailable() < str.length()) {
         blocks_.emplace_back(std::max(DEFAULT_BLOCK_SIZE, str.size()));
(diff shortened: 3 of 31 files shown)


1261385937:
The total cpu improvement : (2650-2400)/2650 = 9.4% 
Exhaustion point is std::vector reallocate.

simple Before:




simple After:

clickhouse/columns/string.h
```diff
@@ -27,6 +27,9 @@ class ColumnFixedString : public Column {
             Append(v);
     }
 
+    /// Increase the capacity of the column for large block insertion.
+    void Reserve(size_t) override {};
```
Enmk, line 31: Please add implementation, it is trivial:
```C++
data_->resize(string_size_ * new_cap, '\0');
```

And also we'll need an explicit counter for a current number of items in the column, rather than implicit (`data_->size() / string_size_`)

clickhouse/columns/tuple.cpp
```diff
@@ -20,6 +20,11 @@ size_t ColumnTuple::TupleSize() const {
     return columns_.size();
 }
 
+void ColumnTuple::Reserve(size_t new_cap)
+{
+    columns_.reserve(new_cap);
```
Enmk, line 25: rather 
```c++
for (auto & column : columns)
  column->Reserve(new_cap);
```

clickhouse/columns/string.cpp
```diff
@@ -190,6 +190,13 @@ ColumnString::ColumnString(std::vector<std::string>&& data)
 ColumnString::~ColumnString()
 {}
 
+void ColumnString::Reserve(size_t new_cap)
+{
+    items_.reserve(new_cap);
+    // 16 is arbitrary number, assumption that string values are about ~256 bytes long.
+    blocks_.reserve(std::max<size_t>(1, new_cap / 16));
```
Enmk, line 197: Why is it different from https://github.com/ClickHouse/clickhouse-cpp/blob/6a8729c5e572146981387275aec62d6b8c0e15ac/clickhouse/columns/string.cpp#L164 ?

Enmk (changes requested):
Please see the comments inline.
Also [add some tests here](https://github.com/ClickHouse/clickhouse-cpp/blob/master/ut/Column_ut.cpp). Some viable test cases:
- Reserve() doesn't change Size() (for both empty and non-empty columns)
- Reserve() on non-empty Column doesn't change stored values.
- Reserve(0) shouldn't crash, also shodun't change stored values.
- How Reserve() and Capacity() work together.

Also please fix the code style, this project uses '[egyptian grackets](https://blog.codinghorror.com/new-programming-jargon/#3)'

clickhouse/columns/lowcardinality.cpp
```diff
@@ -174,6 +174,12 @@ ColumnLowCardinality::ColumnLowCardinality(std::shared_ptr<ColumnNullable> dicti
 ColumnLowCardinality::~ColumnLowCardinality()
 {}
 
+void ColumnLowCardinality::Reserve(size_t new_cap)
+{
+    dictionary_column_->Reserve(new_cap);
```
Enmk, line 179: Here we assume that **ALL** of the new items are unique, wich is quite an uncommon and quite sub-optimal case for `LowCardinality`. 

Maybe estimate dict size as `ln(new_cap) * sqrt(new_cap) + new_cap / 5`, this way we'll get:

| new_cap    | dict's new_cap      | estimated % of unique values |
|------------|---------------------|---------|
| 10 	| 10 	| 100.0  |
| 20 	| 18 	| 90.0  |
| 40 	| 32 	| 80.0  |
| 60 	| 44 	| 73.33  |
| 80 	| 56 	| 70.0  |
| 100 	| 67 	| 67.0  |
| 200 	| 115 	| 57.5  |
| 400 	| 200 	| 50.0  |
| 600 	| 277 	| 46.17  |
| 800 	| 350 	| 43.75  |
| 1000 	| 419 	| 41.9  |
| 2000 	| 740 	| 37.0  |
| 4000 	| 1325 	| 33.12  |
| 6000 	| 1874 	| 31.23  |
| 8000 	| 2404 	| 30.05  |
| 10000 	| 2922 	| 29.22  |
| 20000 	| 5401 	| 27.01  |
| 40000 	| 10120 	| 25.3  |
| 60000 	| 14695 	| 24.49  |
| 80000 	| 19194 	| 23.99  |
| 100000 	| 23641 	| 23.64  |
| 1000000 	| 213816 	| 21.38  |
| 10000000 	| 2050970 	| 20.51  |

clickhouse/columns/tuple.cpp
1261385937: Ok, I made a mistake

clickhouse/columns/string.cpp
1261385937: Usually, 40bytes is a little short, especially for other language (no english). 
I make both them same now.

clickhouse/columns/column.h
```diff
@@ -52,6 +52,9 @@ class Column : public std::enable_shared_from_this<Column> {
     /// Appends content of given column to the end of current one.
     virtual void Append(ColumnRef column) = 0;
 
+    /// Increase the capacity of the column for large block insertion.
+    virtual void Reserve(size_t new_cap) = 0;
```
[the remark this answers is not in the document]
1261385937: If I paln to make a large column about 10w items, i will reserve(10w).
Then i append, if get the 10w count or timeout (maybe 5s), the real insertion will be started.
What i mean is that the reserve size is very exact, more memory needed after reserve is a wrong design.

clickhouse/columns/lowcardinality.cpp
1261385937: Actually, I do not understand the form, so could you  implement ColumnLowCardinality::Reserve ? 🌹

1261385937:
@Enmk thank you for review

clickhouse/columns/lowcardinality.cpp
Enmk: Sure, will do

Enmk (approved):

PR #345, opened 2023-11-15
ColumnArrayT Append will move container data if it is right value

1261385937:
Optimize `ColumnArrayT::Append()` to allow underlying column (e.g. `ColumnString`) to 'steal' values if it has r-value overload of `Append()` to prevent excessive copying.

### DIFF
--- a/clickhouse/columns/array.h
+++ b/clickhouse/columns/array.h
@@ -249,7 +249,7 @@ public:
         if (index >= Size())
             throw ValidationError("ColumnArray row index out of bounds: "
                     + std::to_string(index) + ", max is " + std::to_string(Size()));
-It is actually not, b
+
         return ArrayValueView{typed_nested_data_, GetOffset(index), GetSize(index)};
     }
 
--- a/clickhouse/columns/array.h
+++ b/clickhouse/columns/array.h
@@ -95,7 +95,7 @@ private:
 };
 
 template <typename ColumnType>
-class ColumnArrayT : public ColumnArray {
+class ColumnArrayT final : public ColumnArray {
 public:
     class ArrayValueView;
     using ValueType = ArrayValueView;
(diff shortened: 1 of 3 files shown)


clickhouse/columns/array.h
```diff
@@ -95,7 +95,7 @@ class ColumnArray : public Column {
 };
 
 template <typename ColumnType>
-class ColumnArrayT : public ColumnArray {
+class ColumnArrayT final : public ColumnArray {
```
Enmk, line 98: I assume that `final` is a leftover from a previous PR here. Otherwise, please explain why you need it.

clickhouse/columns/string.cpp
```diff
@@ -232,7 +232,6 @@ void ColumnString::Clear() {
     items_.clear();
     blocks_.clear();
     append_data_.clear();
-    append_data_.shrink_to_fit();
```
Enmk, line 235: Why?

Enmk (changes requested):
Please write a motivation, expected performance effects (if any), etc. for this PR in a description.
Otherwise, it is hard to get the reason behind the proposed changes and whether or not these changes achieve the declared result.

clickhouse/columns/array.h
1261385937: This is a cursoriness. Sorrry

clickhouse/columns/string.cpp
1261385937: Do not do shrink_to_fit for reuse the deque  existing memory.  We just erasure the state.

clickhouse/columns/array.h
```diff
@@ -260,8 +260,17 @@ class ColumnArrayT : public ColumnArray {
...
+    inline void Append(Container&& container) {
+        using container_type = decltype(container);
+        if constexpr (std::is_lvalue_reference_v<container_type> || 
+            std::is_const_v<std::remove_reference_t<container_type>>) {
+            Append(std::begin(container), std::end(container));
+        }
+        else {
+            Append(std::make_move_iterator(std::begin(container)),
+                std::make_move_iterator(std::end(container)));
+            container.clear();
```
Enmk, line 272: IMO this is unnecessary, because after this operation container will be full of moved-out values. There is no way to spoil column data using that container or cause any other harm. If users want to clear container, they can do it after the call, in the user code.

1261385937: why add  container.clear() ?

Enmk (approved):
LGTM

PR #346, opened 2023-11-16
Better estimation for ColumnLowCardinality::Reserve and ColumnString::Reserve

Enmk:
ColumnLowCardinality assumes that not all items are unique, hence dictionary column can be reserved for smaller capacity.

ColumnString now allows to set average value size estimation  in constructor or on existing instance.
If estimation is close to real average value size, then memory pre-allocations are close to optimum.

Added method `MemoryUsage` that provides insight on the amount of RAM used by column instance.


Enmk:
Closing to re-open and to re-trigger CI/CD

PR #355, opened 2024-01-27
Added `GetSize()` const method for FixedStringType

german3w1:
By using ColumnLowCardinalityT<ColumnFixedString>, you lose the ability to conveniently know the fixed size parameter of ColumnFixedString. Adding this method fixes this issue.

### DIFF
--- a/clickhouse/types/types.h
+++ b/clickhouse/types/types.h
@@ -251,6 +251,8 @@ public:
 
     std::string GetName() const { return std::string("FixedString(") + std::to_string(size_) + ")"; }
 
+    inline size_t GetSize() const { return size_; }
+
 private:
     size_t size_;
 };


Enmk:
Hi @german3w1 ! Thank your contribution, could you please add some unit-test for this method? Nothing fancy, since the method itself is pretty straightforward.

Otherwise it looks Ok

german3w1:
I think three unit-tests are enough to prove the method works as expected :)

german3w1:
The building script for Mac with ssl support enabled seems broken to me. Other checks were successful so I think it can be merged now

german3w1:
@Enmk kind ping

Enmk (approved):
LGTM

PR #391, opened 2024-09-13
add clause for inserting queries

### LINKED ISSUE #389: enable async_insert by insert-query
Client insert interface need  support the below statement:
INSERT INTO YourTable SETTINGS async_insert=1, wait_for_async_insert=1 VALUES (...)

1261385937:
INSERT INTO YourTable SETTINGS async_insert=1, wait_for_async_insert=1 VALUES (...)

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -152,7 +152,7 @@ public:
 
     void SendCancel();
 
-    void Insert(const std::string& table_name, const std::string& query_id, const Block& block);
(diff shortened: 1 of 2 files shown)


1261385937:
async inserts only work over the TCP native protocol if the data is sent as SQL text. If you use a client to manage your native protocol inserts actually using the native data format, async inserts are not available.

1261385937:
async inserts only work over the TCP native protocol if the data is sent as SQL text. If you use a client to manage your native protocol inserts actually using the native data format, async inserts are not available.

PR #394, opened 2024-09-20
Parameters in clickhouse-cpp

OlegGalizin:
- **params**
- **Nullable params**

### DIFF
--- a/clickhouse/base/wire_format.cpp
+++ b/clickhouse/base/wire_format.cpp
@@ -173,7 +173,7 @@ void WireFormat::WriteQuotedString(OutputStream& output, std::string_view value)
 }
 
 void WireFormat::WriteParamNullRepresentation(OutputStream& output) {
-    const std::string NULL_REPRESENTATION("'\\\\N'");
+    const std::string NULL_REPRESENTATION(R"('\\N')");
     WriteVarint64(output, NULL_REPRESENTATION.size());
     WriteAll(output, NULL_REPRESENTATION.data(), NULL_REPRESENTATION.size());
 }
--- a/clickhouse/base/wire_format.cpp
+++ b/clickhouse/base/wire_format.cpp
@@ -103,8 +103,9 @@ bool WireFormat::SkipString(InputStream& input) {
 
 inline const char* find_quoted_chars(const char* start, const char* end)
 {
-    static constexpr std::array quoted_chars {'\0', '\b', '\t', '\n', '\'', '\\'};
-    const auto first  = std::find_first_of(start, end, quoted_chars.begin(), quoted_chars.end());
+    static constexpr char quoted_chars[] = {'\0', '\b', '\t', '\n', '\'', '\\'};
+    const auto first  = std::find_first_of(start, end, std::begin(quoted_chars), std::end(quoted_chars));
+
     return (first == end) ? nullptr : first;
 }
 
--- a/clickhouse/base/wire_format.cpp
+++ b/clickhouse/base/wire_format.cpp
@@ -1,3 +1,4 @@
+#include <assert.h>
 #include "wire_format.h"
 
 #include "input.h"
@@ -99,4 +100,76 @@ bool WireFormat::SkipString(InputStream& input) {
     return false;
 }
 
+const char quoted_chars[] = {'\0', '\b', '\t', '\n', '\'', '\\'};
+
+inline const char* find_quoted_chars(const char* start, const char* end) {
+    while (start < end) {
+        char c = *start;
+        for (unsigned i = 0; i < sizeof(quoted_chars); i++) {
+            if (quoted_chars[i] == c) return start;
+        }
+        start++;
+    }
+    return nullptr;
+}
+
+void WireFormat::WriteQuotedString(OutputStream& output, std::string_view value) {
+    auto size               = value.size();
+    const char* start       = value.data();
+    const char* end         = start + size;
+    const char* quoted_char = find_quoted_chars(start, end);
+    if (quoted_char == nullptr) {
+        WriteVarint64(output, size + 2);
+        WriteAll(output, "'", 1);
+        WriteAll(output, start, size);
+        WriteAll(output, "'", 1);
+        return;
+    }
+
+    // calculate quoted chars count
+    int quoted_count             = 1;
+    const char* next_quoted_char = quoted_char + 1;
+    while ((next_quoted_char = find_quoted_chars(next_quoted_char, end))) {
+        quoted_count++;
+        next_quoted_char++;
+    }
+    WriteVarint64(output, size + 2 + 3 * quoted_count);  // length
+
+    WriteAll(output, "'", 1);
+
+    do {
+        auto write_size = quoted_char - start;
+        WriteAll(output, start, write_size);
+        WriteAll(output, "\\", 1);
+        char c = quoted_char[0];
+        switch (c) {
+            case '\0':
+                WriteAll(output, "x00", 3);
+                break;
+            case '\b':
+                WriteAll(output, "x08", 3);
+                break;
+            case '\t':
+                WriteAll(output, "\\\\t", 3);
+                break;
+            case '\n':
+                WriteAll(output, "\\\\\n", 3);
+                break;
+            case '\'':
+                WriteAll(output, "x27", 3);
+                break;
+            case '\\':
+                WriteAll(output, "\\\\\\", 3);
+                break;
+            default:
+                assert(false);
+                WriteAll(output, "x3F", 3); // out ?
+        }
+        start       = quoted_char + 1;
+        quoted_char = find_quoted_chars(start, end);
+    } while (quoted_char);
+
+    WriteAll(output, start, end - start);
+    WriteAll(output, "'", 1);
+}
 }
--- a/clickhouse/base/wire_format.cpp
+++ b/clickhouse/base/wire_format.cpp
@@ -100,12 +100,12 @@ bool WireFormat::SkipString(InputStream& input) {
     return false;
 }
 
-const char quoted_chars[] = {'\0', '\b', '\t', '\n', '\'', '\\'};
+const std::vector<char> quoted_chars = {'\0', '\b', '\t', '\n', '\'', '\\'};
 
 inline const char* find_quoted_chars(const char* start, const char* end) {
     while (start < end) {
         char c = *start;
-        for (unsigned i = 0; i < sizeof(quoted_chars); i++) {
+        for (unsigned i = 0; i < quoted_chars.size(); i++) {
             if (quoted_chars[i] == c) return start;
         }
         start++;
@@ -113,14 +113,7 @@ inline const char* find_quoted_chars(const char* start, const char* end) {
     return nullptr;
 }
 
-void WireFormat::WriteQuotedString(OutputStream& output, std::optional<std::string_view> opt) {
-    if (!opt) //NULL
-    {
-        WriteVarint64(output, 5);
-        WriteAll(output, "'\\\\N'", 5);
-        return;
-    }
-    std::string_view value = *opt;
+void WireFormat::WriteQuotedString(OutputStream& output, std::string_view value) {
     auto size               = value.size();
     const char* start       = value.data();
     const char* end         = start + size;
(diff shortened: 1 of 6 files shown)


OlegGalizin:
What can I do to help you to merge the request?

OlegGalizin:
Is I doing something wrong or nobody has time to merge the request?

1261385937:
@OlegGalizin 
Please sign sign our [Contributor License Agreement](https://cla-assistant.io/ClickHouse/clickhouse-cpp?pullRequest=394)
And we will review your code.

Enmk (changes requested):
Proper high-level description of the added feature is missing. Please add one as a PR description. Having one greatly simplifies review.

Also, could you please create some unit-tests for this functionality? That way there is a chance that nobody is going to break this feature after it is merged. You can add test cases to the `ut/client_ut.cpp` or create a new test unit-test file for parameters specifically.

Enmk:
Hi @OlegGalizin thanks for submitting a PR, please address mentioned issues.

OlegGalizin:
I've add ci test. I did'n found how can I change a PR description. 
The description is
Support query with parameters in clickhouse-cpp

If You can please add the description in proper place.

OlegGalizin:
Please help me run the test

clickhouse/base/wire_format.cpp
```diff
@@ -99,4 +100,83 @@ bool WireFormat::SkipString(InputStream& input) {
     return false;
 }
 
+const char quoted_chars[] = {'\0', '\b', '\t', '\n', '\'', '\\'};
+
+inline const char* find_quoted_chars(const char* start, const char* end) {
+    while (start < end) {
```
Enmk, line 106: Why not [`std::find_first_of`](https://en.cppreference.com/w/cpp/algorithm/find_first_of) ?

clickhouse/base/wire_format.cpp
```diff
@@ -99,4 +100,83 @@ bool WireFormat::SkipString(InputStream& input) {
...
+        for (unsigned i = 0; i < sizeof(quoted_chars); i++) {
+            if (quoted_chars[i] == c) return start;
+        }
+        start++;
+    }
+    return nullptr;
+}
+
+void WireFormat::WriteQuotedString(OutputStream& output, std::optional<std::string_view> opt) {
+    if (!opt) //NULL
```
Enmk, line 117: no need for a `//NULL` comment

clickhouse/base/wire_format.cpp
```diff
@@ -99,4 +100,83 @@ bool WireFormat::SkipString(InputStream& input) {
...
+    {
+        WriteVarint64(output, 5);
```
Enmk, line 119: Please don't use magic constants... it would be nicer to have a constant `NULL_VALUE_REPRESENTATION` pointing to a `"'\\\\N'"` string. And then do `strlen`/`std::string_view::size()`/whatever...

clickhouse/base/wire_format.cpp
```diff
@@ -99,4 +100,83 @@ bool WireFormat::SkipString(InputStream& input) {
...
+                WriteAll(output, "\\\\t", 3);
+                break;
+            case '\n':
+                WriteAll(output, "\\\\\n", 3);
+                break;
+            case '\'':
+                WriteAll(output, "x27", 3);
+                break;
+            case '\\':
+                WriteAll(output, "\\\\\\", 3);
```
Enmk, line 169: Maybe use raw-string literals here? Otherwise it is quite hard to read

clickhouse/base/wire_format.cpp
```diff
@@ -99,4 +100,83 @@ bool WireFormat::SkipString(InputStream& input) {
...
+                break;
+            default:
+                assert(false);
```
Enmk, line 172: is that's an abnormal situation, so client needs to crash in `assert`, or is it a normal situation, and client can skip just write whatever to the stream? You have to decide, and if former, throw an exception. If second, just write stuff to the stream...

clickhouse/base/wire_format.h
```diff
@@ -22,6 +23,7 @@ class WireFormat {
     static void WriteFixed(OutputStream& output, const T& value);
     static void WriteBytes(OutputStream& output, const void* buf, size_t len);
     static void WriteString(OutputStream& output, std::string_view value);
+    static void WriteQuotedString(OutputStream& output, std::optional<std::string_view> value);
```
Enmk, line 26: It looks like this function jams together both "write nullable string" semantics with "write quoted string", is there a way to do those in separate functions?

clickhouse/client.cpp
```diff
@@ -793,6 +813,11 @@ void Client::Impl::SendQuery(const Query& query) {
                 throw UnimplementedError(std::string("Can't send open telemetry tracing context to a server, server version is too old"));
             }
         }
+        if (server_info_.revision >= DBMS_MIN_REVISION_WITH_PARALLEL_REPLICAS) {
+            WireFormat::WriteUInt64(*output_, 0);
```
Enmk, line 817: Please add comments describing what data it skipped / written as `0`

clickhouse/client.cpp
```diff
@@ -817,6 +842,18 @@ void Client::Impl::SendQuery(const Query& query) {
     WireFormat::WriteUInt64(*output_, Stages::Complete);
     WireFormat::WriteUInt64(*output_, compression_);
     WireFormat::WriteString(*output_, query.GetText());
+
+    //Send params after query text
+    if (server_info_.revision >= DBMS_MIN_PROTOCOL_VERSION_WITH_PARAMETERS) {
+        for(const auto& [name, value] : query.GetParams()) {
+            // params is like query settings
+            WireFormat::WriteString(*output_, name);
+            WireFormat::WriteVarint64(*output_, 2); // Custom
```
Enmk, line 851: Please factor out `2` as a named constant.

ut/client_ut.cpp
```diff
@@ -1488,3 +1488,45 @@ TEST(SimpleClientTest, issue_335_reconnects_count) {
...
+        Query query("insert into " + table_name + " values ({id: UInt64}, {name: String})");
+
+        query.SetParam("id", "1").SetParam("name", "NAME");
+        client_->Execute(query);
+
+        query.SetParam("id", "123").SetParam("name", "FromParam");
+        client_->Execute(query);
+
+        query.SetParam("id", "333")
+            .SetParam("name", std::string("A\000A\001A\002A\003A\004A\005A\006A\007A\010A\011A\012A\013A\014A\015A\016A\017A\020A\021A\022A"
```
Enmk, line 1505: It looks like this one doesn't even have any escapable characters in the value... Why?

ut/client_ut.cpp
```diff
@@ -1488,3 +1488,45 @@ TEST(SimpleClientTest, issue_335_reconnects_count) {
...
+                                          "\023A\024A\025A\026A\027A\030A\031A\032A\033A\034"
+                                          "A\035A\036A\037A",
+                                          65));
```
Enmk, line 1508: Again, hardcoded value where you've been using some sort of a constant
```
    const char big_string_value[] = "A\000A\001A\002A\003A\004A\005A\006A\007A\010A\011A\012A\013A\014A\015A\016A\017A\020A\021A\022A"
                                          "\023A\024A\025A\026A\027A\030A\031A\032A\033A\034"
                                          "A\035A\036A\037A";
```

and then do `sizeof(big_string_value) - 1`, accounting for a zero terminator.

ut/client_ut.cpp
```diff
@@ -1488,3 +1488,45 @@ TEST(SimpleClientTest, issue_335_reconnects_count) {
...
+        client_->Execute(query);
+
+        unsigned char big_string[128 - 32];
```
Enmk, line 1511: What does `128 - 32` suppose to mean?

clickhouse/base/wire_format.cpp
OlegGalizin: it more comfortable for me.
I think [std::find_first_of](https://en.cppreference.com/w/cpp/algorithm/find_first_of) needs
on extra compare because it return 'end' on failure
Or after call we needs compare with 'end' that worse then comparing with 0

clickhouse/base/wire_format.cpp
OlegGalizin: It meanings that someone add extra qutated char but did not add it in this switch

clickhouse/client.cpp
OlegGalizin: Im sorry. It something to replication feature. I do not know what it is exactly. You can ask your colleagues

ut/client_ut.cpp
OlegGalizin: on the contrary it have all unprintable characters including \t \n

ut/client_ut.cpp
OlegGalizin: All printable

clickhouse/base/wire_format.cpp
Enmk: Please use functions from standard library whenever possible, instead of inventing new ones with 99% overlapping functionality. "comfortable for me" is not a valid excuse...

clickhouse/base/wire_format.cpp
```diff
@@ -179,4 +171,11 @@ void WireFormat::WriteQuotedString(OutputStream& output, std::optional<std::stri
     WriteAll(output, start, end - start);
     WriteAll(output, "'", 1);
 }
+
+void WireFormat::WriteParamNullRepresentation(OutputStream& output) {
+    const std::string NULL_REPRESENTATION("'\\\\N'");
```
Enmk, line 176: Please us raw string here too

clickhouse/base/wire_format.cpp
OlegGalizin: Wath about  extra compare instruction and slow down the solution?

Enmk: First, it was introduced to be used in a very hot piece of code, which I believe is not the case here.
Second, there is a very high chance that compiler can optimize call `std::find_first_of` (i.e. using SSE) way beyond anything you are trying to gain here.
And last, `find_first_symbols_sse2` supports two modes: "return `nullptr`" and "return `end`" [😉](https://github.com/clickHouse/clickhouse/commit/e22eee80558162bea9a0c7cc5f439f45549dc073).

This looks like you are trying to **prematurely** optimize `O(N + 1)` into `O(N)`, by introducing something that would unnecessary complicate the support.

Enmk:
Please fix the tests

clickhouse/base/wire_format.cpp
```diff
@@ -99,4 +100,82 @@ bool WireFormat::SkipString(InputStream& input) {
     return false;
 }
 
+const std::vector<char> quoted_chars = {'\0', '\b', '\t', '\n', '\'', '\\'};
```
Enmk, line 103: please use `const char quoted_chars[] = ...` instead of `std::vector`, and move this into `find_quoted_chars`, no need to have a global symbol. That would also help performance :)

Enmk:
@OlegGalizin please fix tests

Enmk:
@OlegGalizin ping, `Client/ClientCase.QueryParameters` is failing on almost every platform... 
BTW, to simplify build-test-review loop, you can fix test using your own fork of `clickhouse-ccp`. Runners are provided by github, so it would work on your repo.

OlegGalizin:
> `Client/ClientCase.QueryParameters` is failing on almost every platform

I see. Clickhouse server doesn't support parameters in the test system. 
```
+-----------+
| version() |
+-----------+
|    String |
+-----------+
| 22.3.20.29 |
+-----------+
```
I checked 
ClickHouse 24.7.1.2195 (revision: 54488, git hash: 452d463d77eed30aab18c77933e68316f8c57295
To work with parameters needed revision >= 54459 of ClickHouse Server

OlegGalizin:
> Please fix issues in the PR

I'm sorry. I don't understand you.

Enmk:
> > `Client/ClientCase.QueryParameters` is failing on almost every platform
> 
> @Enmk I see. Clickhouse server doesn't support parameters in the test system.
> 
> ```
> +-----------+
> | version() |
> +-----------+
> |    String |
> +-----------+
> | 22.3.20.29 |
> +-----------+
> ```
> 
> I checked ClickHouse 24.7.1.2195 (revision: 54488, git hash: 452d463d77eed30aab18c77933e68316f8c57295 To work with parameters revision of ClickHouse server must be >= 54459
> 
> Do you want to switch off the test? Will you update the test environment?

Some unit tests are conditionally skipped based on clickhouse version, please check the code and modify your tests accordingly. 

And please either get rid of `find_quoted_chars` or provide a benchmark that shows at least 10% smaller time against `std::find_first_of`. (e.g. using https://quick-bench.com)

OlegGalizin:
> please either get rid of `find_quoted_chars` or provide a benchmark
Please do it himself.

OlegGalizin:
How about merge the PR to master?

OlegGalizin:
@alexey-milovidov Вы можете помочь с этим PR. Уже полгода прошло с момента его создания. То спорили о   двойном цикле. Теперь совсем все заглохло.
С моей стороны весь код готов. Тесты написаны и проходят на сервере с поддержкой параметров.

PR #404, opened 2024-12-13
Changed error detection check when sending data via socket

dngrudin:
Checking the size inequality of the sent data and the input data will not always correctly indicate problems. Sometimes, even for a blocking socket, the send function can return a size smaller than that passed to the function. And only if the function returned the value -1 will it indicate errors.


clickhouse/base/socket.cpp
```diff
@@ -457,11 +457,12 @@ size_t SocketOutput::DoWrite(const void* data, size_t len) {
     static const int flags = 0;
 #endif
 
-    if (::send(s_, (const char*)data, (int)len, flags) != (int)len) {
+    const ssize_t ret = ::send(s_, (const char*)data, (int)len, flags);
```
Enmk, line 460: So is it possible that less data was sent than requested? If yes, what will happen?

dngrudin: I've had this happen (it happens very rarely, so I can't give an example of the sequence of actions that would lead to it). I don't know the exact reasons, since this is not typical for blocking sockets. Perhaps this happens when the server is slow in reading data from the socket compared to writing to the data on the client. I also found an increase in the frequency of such situations when using the 'strace' utility (but perhaps this was due to something else).
If, after the size has returned smaller, resend what was not sent, then all the data will be written to the table.

Enmk: ok, but it seems like resent is not implemented

dngrudin: This is implemented in this place:
```c++
void WireFormat::WriteAll(OutputStream& output, const void* buf, size_t len) {
    const size_t original_len = len;
    const uint8_t* p = static_cast<const uint8_t*>(buf);

    size_t written_previously = 1; // 1 to execute loop at least once
    while (len > 0 && written_previously) {
        written_previously = output.Write(p, len);

        p += written_previously;
        len -= written_previously;
    }

    if (len) {
        throw ProtocolError("Failed to write " + std::to_string(original_len)
                + " bytes, only written " + std::to_string(original_len - len));
    }
}
```
As I understand it, all data is sent through this call.

Enmk (approved):
LGTM

Felixoid:
@socketpair found an issue with it in https://github.com/ClickHouse/clickhouse-cpp/commit/bfd5a077527de083c4db03a16c6683401c396602#r151165234

socketpair:
Actually, according to POSIX, send() may return less than legth. Even zero. It means we should repeat sending unsent part of data in a loop until success or error happens.

Linux may return less in case when socket buffer in kernel is full. Yes, even on blocking socket.

PR #418, opened 2025-04-13
CMake packaging improvements

### LINKED ISSUE #105: Generate CMake config files when installing the project
Please, consider making the corresponding changes in CMake install steps, so that the final distribution (when installed as a separate library) contains CMake- and, possibly, `pkg-config`-related configuration files. This will make consumption of the library easier when it is installed in the system.

See https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html

daneos:
I started working on Debian packaging for the clickhouse-cpp library (not officialy, I need the packages for internal use) and I made some improvements to CMake setup that makes it play nicely with packaging tools.

I thought some of them can be useful for the project.

I tried to split my changes into logical commits so that it's easy to cherry-pick a subset of the changes.

Apparently also closes #105


clickhouse/CMakeLists.txt
```diff
@@ -110,15 +118,23 @@ SET_TARGET_PROPERTIES (clickhouse-cpp-lib
     VERSION ${CLICKHOUSE_CPP_VERSION}
 )
 TARGET_LINK_LIBRARIES (clickhouse-cpp-lib
-    absl::int128
```
xakod, line 113: Why did you mark dependencies private?

daneos: Since all dependencies are static libraries the exported target should not propagate these dependencies further downstream.

xakod: Clickhouse-cpp-lib can be built as shared library

daneos: Yes, this is how I build it for packaging. The dependencies are still static though.

xakod: > Yes, this is how I build it for packaging. The dependencies are still static though.

Dependencies may be shared, for example, conan packaging. In this case, looks like package will be broken.

daneos: The package should not be broken, but the exported CMake target config will be missing dependencies. Is clickhouse-cpp built with conan? I don't see any configuration for it anywhere.

xakod: https://conan.io/center/recipes/clickhouse-cpp?version=2.5.1 and here is an example of usage https://github.com/userver-framework/userver/blob/develop/conanfile.py#L162

daneos: Okay, it seems like conan generates its own CMake config file, so shouldn't be affected.
https://github.com/conan-io/conan-center-index/blob/master/recipes/clickhouse-cpp/all/conanfile.py#L106

I know nothing about conan though, so I'd be grateful if you verify this.

xakod: > it seems like conan generates its own CMake config file, so shouldn't be affected
Okay, I forgot about this. From this point of view, everything is ok. Thank you :)

xakod (approved):

clickhouse/CMakeLists.txt
daneos: There is of course some room for improvement in how I generate the config file (it's probably not really how you are supposed to do it) and it will most likely not cover all possible usecases. I figured out that it might still be useful as is, after all it's an improvement from not having one at all.

daneos:
Updated to current master (sorry it took so long).

Issue #378, opened 2024-07-18
UInt128 is not supported. Only Int128..


socketpair:
https://clickhouse.com/docs/en/sql-reference/data-types/int-uint

Also, I would like to have API where I can pass two 64-bit halves instead of single 128bit value.

socketpair:
Not event speaking of Uint256 values.

Enmk:
Hi @socketpair ! Any contributions are welcome, please submit a patch

PR #424, opened 2025-05-21
Support UInt128 type

pet74alex:
Added UInt128 support. Closes https://github.com/ClickHouse/clickhouse-cpp/issues/378

Tiny backward-incompatible change: UUID type was previously declared by using UInt128 declaration not compatible with the absl one, so the UUID declaration has been changed to remain internally the same but UInt128 has been redeclared by using absl::uint128.


pet74alex:
@Enmk could you please review this and [my previous](https://github.com/ClickHouse/clickhouse-cpp/pull/419)  PRs?

PR #427, opened 2025-06-03
Block API minor improvements

Enmk:
New minor feature.

`Block::Clear()` convenience method that clears out all rows from all columns.
`Block::Reserve()` convenience method that increases capacity of all columns in the block.

Those methods are for the clients that re-use the blocks for multiple batched inserts.

Also `ColumnLowCardinality::Reserve()` now reserves `sqrt(X)` items for dictionary column instead of `X` -- this improves memory usage, since dictionary only holds unique keys of LC columns.


Enmk:
@mshustov kind ping, please review... CI/CD was fixed and all checks are passing: https://github.com/Enmk/clickhouse-cpp/pull/5

Issue #426, opened 2025-05-23
Invalid request sent from clickhouse-cpp


IyeOnline:
I am sending data to a local clickhouse instance using `clickhouse-cpp`. The
program successfully send *some* data, but at some point fails. The point where it fails seems to be random.

My local client then throws a `std::system_error`:

> fail to send 8192 bytes of data: Connection reset by peer

which originates from [socket.cpp:462](https://github.com/ClickHouse/clickhouse-cpp/blob/master/clickhouse/base/socket.cpp#L462)

My clickhouse sever is a local instance simply started via `./clickhouse server`.

Looking into the server's console output/log I can see

<details>
<summary> server log "Unrecognized token: Syntax error"</summary>

```
2025.05.23 14:28:23.925116 [ 46881 ] {40ff087d-cda1-4b93-9c57-e5fbfabfb96a} <Debug> executeQuery: (from 127.0.0.1:50400) INSERT INTO ocsf_dns ( `activity_id`,`activity_name`,`category_uid`,`category_name`,`class_uid`,`class_name`,`severity_id`,`severity`,`type_uid`,`time`,`start_time`,`metadata_log_name`,`metadata_logged_time`,`metadata_product_name`,`metadata_product_vendor_name`,`metadata_product_cpe_name`,`metadata_uid`,`metadata_version`,`answers_rdata`,`answers_ttl`,`query_class`,`query_hostname`,`query_type`,`query_time`,`response_time`,`rcode`,`rcode_id`,`src_endpoint_ip`,`src_endpoint_port`,`dst_endpoint_ip`,`dst_endpoint_port`,`connection_info_direction`,`connection_info_direction_id`,`connection_info_protocol_name`,`connection_info_protocol_num`,`connection_info_protocol_ver_id`,`status`,`status_id`,`unmapped` ) VALUES (stage: Complete)
2025.05.23 14:28:23.931055 [ 46881 ] {40ff087d-cda1-4b93-9c57-e5fbfabfb96a} <Debug> TCPHandler: Processed in 0.006340198 sec.
[...]

PR #428, opened 2025-06-03
Introduce validation to the ColumnArray constructor

IyeOnline:
The ColumnArray constructor would just accept incompatible value- and
offset-arrays and the client would then happily send them, causing
issues in the ClickHouse server.

This introduces validation of the arrays.

Fixes https://github.com/ClickHouse/clickhouse-cpp/issues/426


mshustov (approved):

mshustov (approved):

mshustov (approved):

mshustov:
@IyeOnline could you check the failing tests, please?

Issue #432, opened 2025-07-24
columns/nothing.h is not installed


iso01:
Hi,
Why include file `clickhouse/columns/nothing.h` is not installed?
Can class` clickhouse::ColumnNothing` be used to represent [Nothing](https://clickhouse.com/docs/sql-reference/data-types/special-data-types/nothing) data type?
Best regards,

slabko:
Good catch!
People probably do not use this column very often, which makes sense.

To answer your questions: yes, `clickhouse::ColumnNothing` can be used to represent the Nothing data type. You can also use `clickhouse::ColumnNothing` or even `clickhouse::ColumnNothingT<>` for reading the data. In most cases, I think columns of type Nothing are simply skipped and never even accessed.

Just wondering, what is your use case?

iso01:
No, I am not inserting data using `clickhouse::ColumnNothing`.

I kept the `Nothing` matching to c++ `void` type in some template helper that looks like:

```
template <typename T> struct ch_type {};
template <> struct ch_type<bool> { using type = clickhouse::ColumnUInt8; };
template <> struct ch_type<int8_t> { using type = clickhouse::ColumnInt8; };
template <> struct ch_type<uint8_t> { using type = clickhouse::ColumnUInt8; };
template <> struct ch_type<int16_t> { using type = clickhouse::ColumnInt16; };
...
template <> struct ch_type<std::string> { using type = clickhouse::ColumnString; };
template <> struct ch_type<void> { using type = clickhouse::ColumnNothing; };
```

However when I tried to instantiate my database converter for `void` type, I got compilation error. I started to look around for cpp client examples using `ColumnNothing` and failed. That moment I decided do not spend too much time trying to instantiate converter for `void` type and implemented its use case differently. As you wrote, people do not need this column very often, so I avoided it as well.

I created this issue when realised that even this simple line:
```
template <> struct ch_type<void> { using type = clickhouse::ColumnNothing; };
```
does not compile when headers are taken from clickhouse-cpp client installation area, that seems to be clickhouse cpp client installation bug.

However I am not using it for the moment.

slabko:
I see. Yeah, I would also use `clickhouse::ColumnNothing` in this case.

I updated my original comment and also added `nothing.h` to the list of installed files: https://github.com/ClickHouse/clickhouse-cpp/pull/435. I hope that helps in the future.

PR #435, opened 2025-08-13
Add nothing.h to the list of installed files

slabko:
Fixes https://github.com/ClickHouse/clickhouse-cpp/issues/432

Additionally, add a small test for the NothingColumn type to ensure it is not forgotten.


Enmk:
IMO it would be a good idea to not only add unit-tests, but also installation tests, to prevent the whole class of the problems.

Draft of the steps:
- build `clickhouse-cpp` library
- install it
- ? maybe drop `clickhouse` source directory and build directory to make sure that only recently installed headers/binaries used
- build `clickhouse-cpp-ut` binary against the installed library

PR #441, opened 2025-09-21
clickhouse client move support

ilandry:
As user of the library, the absence of support for move restricts design choices as it forces to use (smart or raw) ptr on Client or to make user classes non-movable too to accommodate.
While this is not dramatic, it is not ideal either. Looking at Client implementation, it has two members:

- a unique_ptr to Impl which by definition should have correct and well defined move
- a const ClientInfo object, as far as I saw all members are movable with non surprising side effect (std::string, std::vector, raw ptr and native integer types), by removing the const qualifier we make sure it is moved rather than copied (which could have unintended consequence for the raw ptr especially).

Hence I do not see large obstacle to make Client movable but happy to have second opinions from regular contributors. This is to kick start some discussions on the topic, I can test the change further on some applications if there is no opposition to the idea.

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -1290,8 +1290,9 @@ Client::Client(const ClientOptions& opts,
 Client::~Client()
 { }
 
-Client::Client(Client&&) = default;
-Client& Client::operator=(Client&&) = default;
+
+Client::Client(Client&&) noexcept = default;
+Client& Client::operator=(Client&&) noexcept = default;
 
 void Client::Execute(const Query& query) {
     impl_->ExecuteQuery(query);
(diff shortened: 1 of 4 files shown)


Enmk (commented):
Please add proper unit-tests (https://github.com/ClickHouse/clickhouse-cpp/blob/master/ut/client_ut.cpp), since `tests/simple/main.cpp` is not built/executed on CI/CD.

Something like:
```
TEST(Client, MoveConstructor) {
...
}


TEST(Client, MoveAssignment) {
...
}
```

ilandry:
Nice, I missed that folder and looked in test only. 
Added unit tests and removed the simple test change.
I used clang-format google style, let me know if you would like me to use a different code format.

slabko (approved):

PR #443, opened 2025-10-30
Add BeginInsert/InsertData/EndInsert flow

theory:
Add a new pattern for "prepared inserts". It works like this:

*   Call `BeginInsert` with an `INSERT` query with optional columns and ending in `VALUES`. No values should be included in the string.
*   It returns a `Block` pre-configured with columns as declared in the `INSERT` statement
*   Add data to the block and periodically call `InsertData` to insert data and clear the block.
*   Call `EndInsert()` or just let the `Client` object go out of scope to signal the server that it's done inserting.

This allows one to send smaller batches of blocks, thereby using less memory, but still in a single ClickHouse `INSERT` operation.

Expected to be useful in the Postgres foreign data wrapper insert API, where multiple rows can be inserted at once but its API handles one-at-a-time insertion. It will also support the FDW COPY API, which can submit huge batches of data to insert, as well.

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -161,6 +161,12 @@ public:
 
     void Insert(const std::string& table_name, const std::string& query_id, const Block& block);
 
+    Block BeginInsert(Query query);
+
+    void InsertData(const Block& block);
+
+    void EndInsert();
+
     void Ping();
 
     void ResetConnection();
@@ -175,6 +181,7 @@ private:
     bool Handshake();
 
     bool ReceivePacket(uint64_t* server_packet = nullptr);
+    bool ReceivePreparePackets(uint64_t* server_packet = nullptr);
 
     void SendQuery(const Query& query, bool finalize = true);
     void FinalizeQuery();
@@ -208,6 +215,7 @@ private:
     }
 
 private:
+    bool inserting;
     /// In case of network errors tries to reconnect to server and
     /// call fuc several times.
     void RetryGuard(std::function<void()> func);
@@ -280,10 +288,15 @@ Client::Impl::Impl(const ClientOptions& opts,
     }
 }
 
-Client::Impl::~Impl()
-{ }
+Client::Impl::~Impl() {
+    // Wrap up an insert if one is in progress.
+    EndInsert();
+}
 
 void Client::Impl::ExecuteQuery(Query query) {
+    if (inserting) {
+        throw ProtocolError("cannot execute query while inserting");
+    }
     EnsureNull en(static_cast<QueryEvents*>(&query), &events_);
 
     if (options_.ping_before_query) {
@@ -299,6 +312,9 @@ void Client::Impl::ExecuteQuery(Query query) {
 
 
 void Client::Impl::SelectWithExternalData(Query query, const ExternalTables& external_tables) {
+    if (inserting) {
+        throw ProtocolError("cannot execute query while inserting");
+    }
     if (server_info_.revision < DBMS_MIN_REVISION_WITH_TEMPORARY_TABLES) {
        throw UnimplementedError("This version of ClickHouse server doesn't support temporary tables");
     }
@@ -362,6 +378,9 @@ std::string NameToQueryString(const std::string &input)
 }
 
 void Client::Impl::Insert(const std::string& table_name, const std::string& query_id, const Block& block) {
+    if (inserting) {
+        throw ProtocolError("cannot execute query while inserting");
+    }
     if (options_.ping_before_query) {
         RetryGuard([this]() { Ping(); });
     }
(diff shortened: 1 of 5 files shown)


clickhouse/client.cpp
```diff
@@ -1085,6 +1179,41 @@ void Client::Impl::RetryGuard(std::function<void()> func) {
...
+    // query columns.
+    auto block = new Block();
+    query.OnData([&block](const Block& b) {
+        for (Block::Iterator bi(b); bi.IsValid(); bi.Next()) {
+            // Create the ClickHouse column type.
+            clickhouse::ColumnRef col = bi.Column();
+            auto chtype = col->Type();
+            if (chtype->GetCode() == Type::LowCardinality) {
+                chtype = col->As<ColumnLowCardinality>()->GetNestedType();
+            }
```
theory: I'm honestly not sure this is the right thing to do. Might one need Type::LowCardonality?

clickhouse/client.cpp
```diff
@@ -161,6 +161,12 @@ class Client::Impl {
 
     void Insert(const std::string& table_name, const std::string& query_id, const Block& block);
 
+    PreparedInsert * PrepareInsert(Query query);
+
+    void FinishInsert();
+
+    void SendData(const Block& block);
```
theory: I had to move this to public so that PreparedInsert can call it. Not in the header file, though, so shouldn't matter.

clickhouse/client.h
```diff
@@ -243,6 +243,21 @@ class Client {
...
     ~Client();
 
+    /**
+     * Prepared insert object.
+     */
+    class PreparedInsert {
+    public:
+        Block * GetBlock();
+        void Execute();
+        // XXX This shouldn't be public.
```
theory: I couldn't figure out how to make this private. Suggestions appreciated.

theory:
Thank you for the design suggestions. I'll work on them this afternoon.

clickhouse/client.h
```diff
@@ -273,6 +273,45 @@ class Client {
...
+     * col1.Append(3);
+     * col2.Append("amos");
+     *
+     * // Send it and finish.
+     * client->EndInsert(*block);
+     * \endcode
+     */
+    std::unique_ptr<Block> BeginInsert(const std::string& query);
+    std::unique_ptr<Block> BeginInsert(const std::string& query, const std::string& query_id);
+    void InsertData(Block& block);
```
theory: Holler if you'd rather pass a `std::unique_ptr<Block>`. Seems okay to me to pass a `*block` instead, but I'm not yet up to snuff on idiomatic C++.

theory: But `InsertData` does modify the block, by design:

```cpp
void Client::Impl::InsertData(Block& block) {
    assert(inserting);
    block.RefreshRowCount();
    SendData(block);
    block.Clear();
}
```

Would you rather that refreshing the count and clearing be done by the caller?

theory: Switched to returning a `Block` in 0a3da16, and also moved the docs to the README. [Diff](https://github.com/ClickHouse/clickhouse-cpp/compare/fb148753cea70e80d0af952ce7b91a5fd01954c6..0a3da16ed87ce0f647b16e43ed8af56f68064242).

theory: Does the caller need to `delete` the returned `Block`? Seems like yes, since IIUC `Block()` creates it on the heap, yes?

clickhouse/client.cpp
```diff
@@ -280,10 +288,15 @@ Client::Impl::Impl(const ClientOptions& opts,
     }
 }
 
-Client::Impl::~Impl()
-{ }
+Client::Impl::~Impl() {
+    // Wrap up an insert if one is in progress.
+    EndInsert();
```
madshell: Not sure this is the right place but it seems this is dangerous in the situation where the database stops when inserting a batch of records.

With v2.6.0 my app correctly recovers if Clickhouse stops in the middle of an insert but after this was added to the destructor, it crashes with this stack:

```
terminate called after throwing an instance of 'std::system_error'
  what(): fail to send 38 bytes of data: Broken pipe
#10 0x000055555561b866 in clickhouse::Client::Impl::EndInsert()
#11 0x0000555555c4823c in clickhouse::Client::Impl::~Impl()
#12 0x0000555555c4857b in clickhouse::Client::~Client()
```

I reproduce by issuing this simple command during a big insert loop:
```
service clickhouse-server stop
```

I think we need at least a try/catch around EndInsert() in the destructor to mitigate the issue but I'm no C++ expert.

slabko: Hi @madshell 

That’s a very good point! It really does make it difficult to handle these kinds of exceptions (and can create a bunch of problems). Putting a try/catch around EndInsert() is probably the best solution, especially since no data loss is involved.

Thank you very much for pointing this out!

slabko: https://github.com/ClickHouse/clickhouse-cpp/pull/455

Issue #284, opened 2023-02-12
abseil library isn't installed resulting in compilation error


pkoosha:
I did
``` git clone --recursive https://github.com/ClickHouse/clickhouse-cpp.git```
then
``` 
make -DBUILD_TESTS=ON -DCMAKE_CXX_FLAGS="-std=c++17" -DCMAKE_CXX_COMPILER=/opt/gcc/7.5.0/bin/g++ -DCMAKE_C_COMPILER=/opt/gcc/7.5.0/bin/gcc -DCMAKE_EXE_LINKER_FLAGS="-L/opt/gcc/7.5.0/lib64" -DCMAKE_INSTALL_PREFIX:PATH=/home/kousha.2/projects/clickhouse-cpp/build/install-dir ..

-- The C compiler identification is GNU 7.5.0
-- The CXX compiler identification is GNU 7.5.0
-- Check for working C compiler: /opt/gcc/7.5.0/bin/gcc
-- Check for working C compiler: /opt/gcc/7.5.0/bin/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /opt/gcc/7.5.0/bin/g++
-- Check for working CXX compiler: /opt/gcc/7.5.0/bin/g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /home/kousha.2/projects/clickhouse-cpp/build
(base) kousha.2@fpga01:~/projects/clickhouse-cpp/build (master)$ make install
Scanning dependencies of target absl_int128
[  1%] Building CXX object contrib/absl/absl/CMakeFiles/absl_int128.dir/numeric/int128.cc.o
[  3%] Linking CXX static library libabsl_int128.a
[  3%] Built target absl_int128
Scanning dependencies of target lz4
[  4%] Building C object contrib/lz4/lz4/CMakeFiles/lz4.dir/lz4.c.o
[  6%] Building C object contrib/lz4/lz4/CMakeFiles/lz4.dir/lz4hc.c.o
[  7%] Linking C static library liblz4.a
[  7%] Built target lz4
Scanning dependencies of target cityhash
[...]

PR #447, opened 2025-11-11
fix abseil includes

jeoseo:
See https://github.com/ClickHouse/clickhouse-cpp/issues/284


slabko:

PR #448, opened 2025-11-13
Clean up and simplify Begin/EndInsert

slabko:
- Renamed `InsertData` to `SendInsertBlock`, to prevent it from appearing first when searching for `Insert...`
- Removed `EndInsert` with a block. Instead, users should now use `SendInsertBlock` followed by `EndInsert`.
- Removed `ReceivePreparePackets` and replaced it with a call to `ReceivePacket`, similar to what is done in `Insert`.
- Renamed `inserting` to `inserting_` to follow the member variable naming convention.
- Replaced `ProtocolError` with `ValidationError` for the situation when calling other operations is not allowed between `BeginInsert` and `EndInsert`
- Move the `BeginInsert` body to match the order of the class declaration, placing it between `Insert` and `SendInsertData`.

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -417,6 +417,39 @@ void Client::Impl::Insert(const std::string& table_name, const std::string& quer
     throw ProtocolError("fail to receive data packet");
 }
 
+Block Client::Impl::BeginInsert(Query query) {
+    if (inserting_) {
+        throw ValidationError("cannot execute query while inserting");
+    }
+
+    EnsureNull en(static_cast<QueryEvents*>(&query), &events_);
+
+    if (options_.ping_before_query) {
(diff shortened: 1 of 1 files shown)


clickhouse/client.cpp
```diff
@@ -1194,15 +1122,18 @@ void Client::Impl::RetryGuard(std::function<void()> func) {
 }
 
 Block Client::Impl::BeginInsert(Query query) {
-    if (inserting) {
+    if (inserting_) {
         throw ProtocolError("cannot execute query while inserting");
```
theory: Do you want to change this one to `ValidationError()`, too?

slabko: Oops, my bad, yeah, meant to that too!

PR #449, opened 2025-11-18
Start adding Bool type

theory:
Try implementing it as a number, just like `UInt8`. One test fails:

```
/Users/david/dev/clickhouse/clickhouse-cpp/ut/CreateColumnByType_ut.cpp:72: Failure
Expected equality of these values:
  col->GetType().GetName()
    Which is: "UInt8"
  GetParam()
    Which is: "Bool"
```

It looks as though `CreateColumnByType`, even though passed a `Bool`, creates a `UInt8`. I tried instead configuring it with

```patch
--- a/clickhouse/columns/numeric.cpp
+++ b/clickhouse/columns/numeric.cpp
@@ -108,6 +108,7 @@ ItemView ColumnVector<T>::GetItem(size_t index) const  {
     return ItemView{type_->GetCode(), data_[index]};
 }
 
+template class ColumnVector<bool>;
 template class ColumnVector<int8_t>;
 template class ColumnVector<int16_t>;
 template class ColumnVector<int32_t>;
--- a/clickhouse/columns/numeric.h
+++ b/clickhouse/columns/numeric.h
@@ -70,7 +70,7 @@ using Int128 = absl::int128;
 using UInt128 = absl::uint128;
 using Int64 = int64_t;
 
-using ColumnBool    = ColumnVector<uint8_t>;
+using ColumnBool    = ColumnVector<bool>;
 using ColumnUInt8   = ColumnVector<uint8_t>;
 using ColumnUInt16  = ColumnVector<uint16_t>;
 using ColumnUInt32  = ColumnVector<uint32_t>;
```

But that won't even compile, presumably because a vector of bools is a whole different thing than a vector of numbers.

Not a blocker for my project, as the old mapping to UInt8 seems to work well enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ C++ client for [ClickHouse](https://clickhouse.com/).
 * String
 * LowCardinality(String) or LowCardinality(FixedString(N))
 * Tuple
-* UInt8, UInt16, UInt32, UInt64, Int8, Int16, Int32, Int64
+* UInt8, UInt16, UInt32, UInt64, Int8, Int16, Int32, Int64, Bool
 * UInt128, Int128
 * UUID
 * Map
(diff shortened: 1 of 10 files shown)


theory:
Replaced by #492.

PR #450, opened 2025-11-19
Add JSON support

theory:
Implemented the same as String, which seems to work fine except for round-tripping. I assume it's not actually a string in the protocol so it will need a different format. But hopefully the rest of the infrastructure for it is helpful.

Not a blocker for my project; JSON works fine for now with the HTTP interface.


theory:
Closing in favor of #490.

PR #453, opened 2025-12-03
Allow either *char or Query as BeginInsert arg

theory:
To allow the user to configure other bits of the query, like settings.


serprex (approved):

theory:
We no longer need this for clickhouse-cpp, but leaving it as it may be useful to others.

slabko:
Already included in the library

PR #454, opened 2025-12-09
Keep absl API available in the shared library

slabko:


slabko:

PR #471, opened 2026-04-02
Add a pull-style streaming select API

iskakaushik:
pg_clickhouse needs to consume select results one block at a time, but clickhouse-cpp only exposes a callback-driven select path today. That forces downstream users to layer coroutines or connection resets on top of the client when they need pull-style iteration.

Add BeginSelect(), ReceiveSelectBlock(), and EndSelect() to mirror the existing multi-step insert workflow. The implementation reuses the existing query and packet handling code, keeps Query callbacks active for progress, profile, and log packets, and drains canceled queries so connections remain reusable.

Add integration tests that cover full streaming iteration, preserved Query callbacks, early cleanup, end-of-stream reuse, and exception cleanup with subsequent reuse.

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -1,4 +1,5 @@
 #include "client.h"
+#include "clickhouse/error_codes.h"
 #include "clickhouse/version.h"
 #include "protocol.h"
 
@@ -157,6 +158,12 @@ public:
 
     void SelectWithExternalData(Query query, const ExternalTables& external_tables);
 
+    void BeginSelect(const Query& query);
+
+    std::optional<Block> ReceiveSelectBlock();
+
+    void EndSelect();
+
     void SendCancel();
 
     void Insert(const std::string& table_name, const std::string& query_id, const Block& block);
@@ -208,6 +215,14 @@ private:
 
     void InitializeStreams(std::unique_ptr<SocketBase>&& socket);
 
+    void EnsureIdle(const char* action) const;
+
+    uint64_t DrainQueryResponse(const char* context);
+
+    void ResetSelectState();
+
+    std::optional<Block> TakeSelectBlock();
+
     inline size_t GetConnectionAttempts() const
     {
         return options_.endpoints.size() * options_.send_retries;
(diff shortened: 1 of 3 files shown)


slabko:
@iskakaushik master seem to build fine, can you take a look at CI errors.

clickhouse/client.cpp
```diff
@@ -258,7 +273,12 @@ class Client::Impl {
 
     ServerInfo server_info_;
 
-    bool inserting_;
+    bool inserting_ = false;
+    bool selecting_ = false;
+    bool discarding_select_data_ = false;
```
slabko, line 278: I believe this will not be needed if after proper synchronous version of `ReceivePacket` is implemented.

clickhouse/client.cpp
```diff
@@ -258,7 +273,12 @@ class Client::Impl {
 
     ServerInfo server_info_;
 
-    bool inserting_;
+    bool inserting_ = false;
```
slabko, line 276: A bunch mutually exclusive of bools look time like a good candidate for a enum.

slabko (changes requested):
After some back and forth, implementing a `ReceivePacket` that only parses a single packet and returns a tagged enum (`std::variant`) does not seem particularly complex. The current `ReceivePacket` can then be built on top of it.

This gives full control over the `ExecuteQuery` loop, allowing a synchronous implementation without workarounds. In other words, the library can become synchronous while still preserving the asynchronous (callback-based) API.

This also enables synchronous handling of other events.

Finally, `clickhouse::Query` callbacks should not be ignored in synchronous mode—they can still be invoked, making the implementation complete.

There will be some extra work to be done, but it will be clear what after experimenting more with synchronous version of `ReceivePacket`

slabko:
@iskakaushik 
Here ReceivePacket now returns the data that it has received: https://github.com/ClickHouse/clickhouse-cpp/pull/474

Regarding API naming, for a pull-based API I would use `BeginSelect` and `SelectNext`, which returns the next block (and still passes it through callbacks). Based on that, I could implement `SelectAll`, which passes all blocks through the callbacks. After that, I could re-implement `Select` by calling `PrepareSelect` and `SelectAll`. And here we are, with `Select` implemented in terms of the new pull-based `ReceivePacket`, while still preserving all callbacks embedded in `Query`.

Based on that it is relatively easy to return data to the caller directly and allow interactive control over the loop.

This is where I would start.

mshustov:
implemented in https://github.com/ClickHouse/clickhouse-cpp/pull/474

PR #483, opened 2026-04-21
Create cross-repo-bug-relay.yml

alex-clickhouse:
Sets up new issue relay for cross-repo investigation.

If you'd rather trigger this on labeling an issue a bug we can do this instead:

```
  on:
    issues:
      types: [labeled]

  permissions: {}

  jobs:
    relay:
      if: github.event.label.name == 'bug'
      uses: ClickHouse/integrations-shared-workflows/.github/workflows/cross-repo-bug-relay.yml@main
      secrets: inherit
```

But there's a triage before the cross-repo investigation that filters out irrelevant issues.

### DIFF
--- a/.github/workflows/cross-repo-bug-relay.yml
+++ b/.github/workflows/cross-repo-bug-relay.yml
@@ -1,0 +1,17 @@
+name: Relay bugs for cross-repo investigation
+
+# Relays newly-opened issues to ClickHouse/integrations-ai-playground for
+# cross-repo investigation.
+
+on:
+  issues:
+    types: [opened]
+
+permissions: {}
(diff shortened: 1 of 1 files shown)


slabko:
@alex-clickhouse, How does it work when I create a new issue with the label already set during creation? I suspect GitHub will miss the _labeled_ event altogether and only emit _created_.

alex-clickhouse:
As-is it triggers on issue creation regardless of label.

It's a good q whether the labeled event would trigger on creation. The docs are unclear and I didn't find anything, would have to test.

slabko (approved):

PR #485, opened 2026-04-21
Add Tuple field names and Bool Type

### LINKED ISSUE #476: Add Bool type support (distinct from UInt8)
ClickHouse has a Bool type distinct from UInt8. clickhouse-cpp currently has no Type::Code::Bool — Bool columns are silently handled as UInt8. pg_clickhouse maps Bool to PostgreSQL BOOLEAN at the schema level, but the binary driver has no way to distinguish Bool from UInt8 at the wire level, which causes type mismatches during reads.

### LINKED ISSUE #484: Provide nested field names for Tuple
Tuple fields can be named in ClickHouse, but the client library does not provide access to the names through `TupleType`.

This makes it necessary to run a separate `DESCRIBE TABLE` query and then manually parse the result if one needs the field names, which is highly inconvenient.

IyeOnline:
This resolves two issues I just ran into when writing an integration to fetch data from ClickHouse.

* A strong Bool type to make Bool Columns truly distinct from UInt8 columns.
* A way to obtain the names for the fields of a Tuple.

Both changes are rather simple and self-contained in indivudual commits. Each adds a test for the newly added functionality.

- Fixes https://github.com/ClickHouse/clickhouse-cpp/issues/476
- Fixes https://github.com/ClickHouse/clickhouse-cpp/issues/484

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -278,3 +278,6 @@ BUCKAROO_DEPS
 # Vim
 *.swp
 *.swo
+
+# clangd cache
+/.cache/clangd
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ C++ client for [ClickHouse](https://clickhouse.com/).
 ## Supported data types
 
 * Array(T)
+* Bool
 * Date
 * DateTime, DateTime64
 * DateTime([timezone]), DateTime64(N, [timezone])
@@ -256,5 +257,3 @@ client.Insert("default.test", block);
 ```sql
 ALTER USER insert_account SETTINGS async_insert=1,wait_for_async_insert=1,async_insert_use_adaptive_busy_timeout=0,async_insert_busy_timeout_ms=5000,async_insert_max_data_size=104857600
 ```
-
-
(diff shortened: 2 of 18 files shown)


IyeOnline:
@mshustov I thought about this a bit more over the evening and realized that the addition of `Bool` is an potential API break towards consumers of the library. For example if they expect `Bool`'s to be `Uint8`'s or if they think they covered all types, but now a new one was added.

I don't know what your policy is for such a potentially breaking change. I saw that 2.6.0 added new types, so I assume it would be fine. Still I think this deserves some consideration, given that `Bool` is going to be a fairly common type.

slabko:
Hi @IyeOnline
Thank you very much for your contribution - this absolutely makes sense for the library. However, could you please create two separate PRs, since they are two very different pieces of functionality, so we can have separate discussions and merge them separately?

slabko:
@IyeOnline To avoid breaking ABI, I suggest introducing a compiler definition, `CH_MAP_BOOL_TO_UINT8`, enabled by default and phased out in future versions.

IyeOnline:
Done and Done. Closing this in favor of the two separate PRs

- https://github.com/ClickHouse/clickhouse-cpp/pull/491
- https://github.com/ClickHouse/clickhouse-cpp/pull/492

PR #491, opened 2026-04-29
Add Tuple Element Names to `TupleType`

### LINKED ISSUE #484: Provide nested field names for Tuple

IyeOnline:
This extends the type parser to support an element name alongside the
element type and wires this through to make the field names in `TupleType`
available in the C++ API.

- Fixes https://github.com/ClickHouse/clickhouse-cpp/issues/484

### DIFF
--- a/clickhouse/columns/tuple.cpp
+++ b/clickhouse/columns/tuple.cpp
@@ -16,6 +16,13 @@ ColumnTuple::ColumnTuple(const std::vector<ColumnRef>& columns)
 {
 }
 
+ColumnTuple::ColumnTuple(const std::vector<ColumnRef>& columns,
+                         std::vector<std::string> names)
+    : Column(Type::CreateTuple(CollectTypes(columns), std::move(names)))
+    , columns_(columns)
+{
+}
+
 size_t ColumnTuple::TupleSize() const {
     return columns_.size();
 }
@@ -48,7 +55,11 @@ ColumnRef ColumnTuple::Slice(size_t begin, size_t len) const {
         sliced_columns.push_back(column->Slice(begin, len));
     }
 
-    return std::make_shared<ColumnTuple>(sliced_columns);
+    const auto& names = this->Type()->As<TupleType>()->GetItemNames();
+    if (names.empty()) {
+        return std::make_shared<ColumnTuple>(sliced_columns);
+    }
+    return std::make_shared<ColumnTuple>(sliced_columns, names);
 }
 
 ColumnRef ColumnTuple::CloneEmpty() const {
@@ -59,7 +70,11 @@ ColumnRef ColumnTuple::CloneEmpty() const {
         result_columns.push_back(column->CloneEmpty());
     }
 
-    return std::make_shared<ColumnTuple>(result_columns);
+    const auto& names = this->Type()->As<TupleType>()->GetItemNames();
+    if (names.empty()) {
+        return std::make_shared<ColumnTuple>(result_columns);
+    }
+    return std::make_shared<ColumnTuple>(result_columns, names);
 }
 
 bool ColumnTuple::LoadPrefix(InputStream* input, size_t rows) {
--- a/clickhouse/columns/tuple.cpp
+++ b/clickhouse/columns/tuple.cpp
@@ -1,13 +1,40 @@
 #include "tuple.h"
 
 namespace clickhouse {
+namespace {
 
-static std::vector<TypeRef> CollectTypes(const std::vector<ColumnRef>& columns) {
+std::vector<TypeRef> CollectTypes(const std::vector<ColumnRef>& columns) {
     std::vector<TypeRef> types;
     for (const auto& col : columns) {
         types.push_back(col->Type());
     }
     return types;
+}
+
+/// Tuple types can be appended if they have the same shape.
+bool CanAppendType(const TypeRef& destination_type, const TypeRef& source_type) {
+    if (destination_type->GetCode() != Type::Tuple || source_type->GetCode() != Type::Tuple) {
+        return destination_type->IsEqual(source_type);
+    }
+
+    const auto* destination_tuple = destination_type->As<TupleType>();
+    const auto* source_tuple      = source_type->As<TupleType>();
+
+    const auto destination_item_types = destination_tuple->GetTupleType();
+    const auto source_item_types      = source_tuple->GetTupleType();
+    if (destination_item_types.size() != source_item_types.size()) {
+        return false;
+    }
+
+    for (size_t i = 0; i < destination_item_types.size(); ++i) {
+        if (!CanAppendType(destination_item_types[i], source_item_types[i])) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
 }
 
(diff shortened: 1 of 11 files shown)


clickhouse/types/types.h
```diff
@@ -126,6 +126,9 @@ class Type {
 
     static TypeRef CreateTuple(const std::vector<TypeRef>& item_types);
 
+    static TypeRef CreateTuple(const std::vector<TypeRef>& item_types,
+                               std::vector<std::string> item_names);
+
```
slabko, line 129: Can we replace the function above with this one and set item_names to {} by default, or am I missing something?

clickhouse/types/type_parser.h
```diff
@@ -31,6 +31,9 @@ struct TypeAst {
     /// Type's name.
     /// Need to cache TypeAst, so can't use StringView for name.
     std::string name;
+    /// Name of this element inside its parent (e.g. field name inside a named
+    /// Tuple). Empty for unnamed elements.
+    std::string element_name;
```
slabko, line 36: I think it would make more sense to add a vector here, called `all_names`, store both the name and the type in it, and keep the name field set to the last value as it currently works.

```
/// Type's name.
/// Need to cache TypeAst, so can't use StringView for name.
std::string name;
/// List of all names assigned to the element, for example 
/// named tuples, have name and type, `Tuple(i Int64, s String)`.
std::vector<std::string> all_names

File: clickhouse/types/types.cpp [hunk 3]
```diff
@@ -456,13 +466,30 @@ LowCardinalityType::~LowCardinalityType() {
 
 std::string TupleType::GetName() const {
     std::string result("Tuple(");
+    bool has_complete_names = item_names_.size() == item_types_.size();
+    if (has_complete_names) {
```
slabko: This check should be performed during type construction to ensure that any function using it is guaranteed to have either an empty `item_names_` or the same number of elements as `item_types_`.

clickhouse/types/types.cpp
```diff
@@ -456,13 +466,30 @@ LowCardinalityType::~LowCardinalityType() {
 
 std::string TupleType::GetName() const {
```
slabko, line 466: There is an interesting side effect associated with the updated type name. I cannot use `ColumnTuple::Append(ColumnRef column)` to append columns to a named tuple that has the same structure but a different name, even though, from my point of view, nothing should prevent the user from doing so.

slabko (dismissed):
@IyeOnline

Thank you very much for your PR. This absolutely makes sense and is well implemented. I’ve added a couple of points; they’re not critical, so feel free to ignore them if you don’t have time to address them. You can go ahead and merge, and I’ll take care of the remaining required changes.

clickhouse/types/type_parser.h
IyeOnline: Personally I dont really like that, because it
* Creates two spots containing the same information
* Does not make it clear through the type-system/name that the entries in `all_names` are rather different entities (field name, vs type name)

clickhouse/types/types.cpp
IyeOnline: I decided to soften this in `Append` by checking for shape similarity and "compatible" names.

clickhouse/columns/tuple.cpp
```diff
@@ -10,6 +10,46 @@ static std::vector<TypeRef> CollectTypes(const std::vector<ColumnRef>& columns)
...
+    const auto source_item_types      = source_tuple->GetTupleType();
+    if (destination_item_types.size() != source_item_types.size()) {
+        return false;
+    }
+
+    const auto& destination_names = source_tuple->GetItemNames();
+    const auto& source_names      = source_tuple->GetItemNames();
+    if (!HasNoFieldNames(source_names) && destination_names != source_names) {
+        return false;
+    }
```
slabko: I might be missing something, but don’t we avoid comparing names? Does it matter what the names are if the types are compatible?

clickhouse/columns/tuple.cpp
```diff
@@ -10,6 +10,46 @@ static std::vector<TypeRef> CollectTypes(const std::vector<ColumnRef>& columns)
...
+        if (!name.empty()) {
+            return false;
+        }
+    }
+    return true;
+}
+
+/// Tuple types can be appended if they have the same shape and
+/// the to-be-appended type has the same names or is entirely unnamed.
+static bool CanAppendType(const TypeRef& destination_type, const TypeRef& source_type) {
```
slabko: Not a big deal at all, but it might be worth considering making this a member function. It could help avoid any confusion between the source and destination, and would keep the interface to a single parameter - the type to append. Also it looks like the project tends to use anonymous namespaces rather than static for internal linkage.

clickhouse/types/types.cpp
```diff
@@ -466,13 +460,11 @@ LowCardinalityType::~LowCardinalityType() {
...
-    if (has_complete_names) {
-        for (const auto& item_name : item_names_) {
-            if (item_name.empty()) {
-                has_complete_names = false;
-                break;
-            }
+    bool has_complete_names = !item_names_.empty();
+    for (const auto& item_name : item_names_) {
+        if (item_name.empty()) {
+            has_complete_names = false;
```
slabko: I might be misunderstanding something here, but if we ensure during construction that `item_names_` is either empty or contains valid names for each type, would this additional check still be necessary? It seems like `bool has_complete_names = !item_names_.empty()` is sufficient.

clickhouse/columns/tuple.cpp
IyeOnline: IMO it does matter. In my book tuples with different field names are different types. I would consider it a mistake to try and append them.

If a user really wants to append columns of tuples, they can make sure that either the appended column is unnamed or has the same name.

slabko: I’m not sure I fully follow. This sounds like an artificial constraint to me. Could you clarify why adding a column with matching types but different names would be an issue, especially since it was possible before this PR?

clickhouse/columns/tuple.cpp
IyeOnline: The problem is that we need to support recursion, so unless we extend the generic `Type` or `TupleType` interface, this cant work.

I don't want to introduce a concept of "type similarity" in this PR, so I think its best to just stay with the free function.

clickhouse/columns/tuple.cpp
IyeOnline: It was "possible" before this PR insofar that there simply were no names at all.

I am probably looking at this too much from our own type system though. I just played around with ClickHouse SQL a bit and it seems most operations do not care about the name at all.

Hence I now think its OK to append columns of mismatching names 👍

IyeOnline:
@slabko Can you take another look 🙏

slabko:
@IyeOnline Sorry for the delay. I’ll take a close look first thing next week, but so far everything looks good and I’ll probably merge it.

slabko (approved):

PR #492, opened 2026-04-29
Introduce `Bool` Type

### LINKED ISSUE #476: Add Bool type support (distinct from UInt8)

IyeOnline:
Previously the library implicitly converted bool columns to `Uint8`,
loosing type information.

This commit introduces a "strong type" for `Bool` that is distinct from
`bool`. This allows complete re-use of `ColumnVector` without triggering
the `std::vector<bool>` specialization.

This new type is guarded behind a CMake variable/preprocessor define
`CH_MAP_BOOL_TO_UINT8`, which is enabled by default in order to keep
compatibility for library users. In the future the default should be
switched first and the variable then removed later.

- Fixes https://github.com/ClickHouse/clickhouse-cpp/issues/476

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -98,6 +98,7 @@ jobs:
             -D CMAKE_C_COMPILER=${{matrix.C_COMPILER}} \
             -D CMAKE_CXX_COMPILER=${{matrix.CXX_COMPILER}} \
             -D CMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} \
+            -D CH_MAP_BOOL_TO_UINT8=OFF \
             -D BUILD_TESTS=ON \
             ${{matrix.SSL_CMAKE_OPTION}} \
             ${{matrix.DEPENDENCIES_CMAKE_OPTIONS}} \
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -48,6 +48,7 @@ jobs:
         run: |
           cmake \
             -D CMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} \
+            -D CH_MAP_BOOL_TO_UINT8=OFF \
             -D BUILD_TESTS=ON \
             ${{matrix.SSL_CMAKE_OPTION}} \
             -S ${{github.workspace}} \
--- a/.github/workflows/windows_mingw.yml
+++ b/.github/workflows/windows_mingw.yml
@@ -52,7 +52,7 @@ jobs:
             tar
 
       - name: Configure CMake
-        run: cmake -B build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DBUILD_TESTS=ON -DCHECK_VERSION=OFF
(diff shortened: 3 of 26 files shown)


slabko:
@IyeOnline, I think I’m missing something here, but how am I supposed to use it for inserting values? The obvious way of using the actual type doesn’t look right to me:

```
col_bool->Append(clickhouse::Bool::true_);
col_bool->Append(clickhouse::Bool::false_);
```

This doesn’t look any better either:

```
col_bool->Append(static_cast<clickhouse::Bool>(true));
col_bool->Append(static_cast<clickhouse::Bool>(false));
```

I was wondering whether you considered simply subclassing `ColumnVector<uint8_t>`?

IyeOnline:
@slabko You can do `col_bool->Append(Bool{true});`, potentially with a `using ::clickhouse::Bool` in scope - which really isnt much better.

I only tested getting values out on my side (because thats what I need it for) and there the `clickhouse::Bool -> bool` conversion is implicit. I actually intended for this to work without any extra specifications and it clearly does not. :|

I did think about sub-classing a uint8 column originally as a means of having seamless compatibility but discarded that idea as it would still clash with the `Type::Code`.

Further, I think for correctness you would want the ColumnBool to only privately inherit `ColumnVector<uint8_t>` as its only an implementation detail. This made it not very appealing, since you then need to introduce virtual inheritance and a bunch of `using` declarations to get public members form the private base back.

But looking at the `Append` issue (which I missed/thought would work), I think that is indeed the way to go.

IyeOnline:
@slabko I went with composition instead of inheritance, as I think it would be preferable to not be able to cast a column with `type == Bool` to UInt8, but I dont have a strong opinion on this. Private inheritance would have required making every inheritance `virtual`, which would have required larger changes (e.g. not being able to `static_cast` in a bunch of places).

slabko:
@IyeOnline I think composition is even better and provides much more flexibility.

~Do you have a chance to look at the MSVC build? It seem to be failing on CI.~
Nevermind, fixed it.

slabko:
rebased the PR on master

slabko:
@IyeOnline Do we need to hide everything Bool-related when `CH_MAP_BOOL_TO_UINT8` is enabled? It seems like the only place where this definition can actually be used is in `type_parser.cpp:32` were we define mapping from ClickHouse's `Bool` to either `Type::UInt8` or `Type::Bool`. I couldn't find anything else that would break or behave unexpectedly if remove this macro check in other places (except tests)

slabko:
@IyeOnline Can you please take a look at my changes, especially the reduction in the use of `CH_MAP_BOOL_TO_UINT8`? I have a feeling that I might be missing something here.

IyeOnline:
@slabko I think you are correct that there is no reason to hide the type entirely. When creating columns in user code, one can use either column type just fine. All that matters for compatibility is what the library produces when getting blocks from the server and that is just that one mapping site.

I made a small edit to the readme to clarify this.

slabko (approved):

PR #499, opened 2026-05-20
Fix parsing of quoted identifiers

IyeOnline:
Previously a type such as

    Tuple( `a.b` Int8)

would fail to parse. This would cause `Client::Impl::ReadBlock` to throw
an "unsupported column type" exception, even though this is a valid
ClickHouse type and supported by the library.

This fix introduces a new token type for quoted identifiers and uses this
tuples. A dedicated token type seems a semantically cleaner choice, because
formally there could be cases where `Name` would be accepted, but a
quoted identifier would not be allowed.

This is effectively a followup to https://github.com/ClickHouse/clickhouse-cpp/pull/491

### DIFF
--- a/clickhouse/types/type_parser.cpp
+++ b/clickhouse/types/type_parser.cpp
@@ -266,38 +266,12 @@ TypeParser::Token TypeParser::NextToken() {
             {
                 const auto quote = *cur_;
                 ++cur_;
-                const auto start = cur_;
-
-                // Fast path: scan for the closing quote with no escape
-                // sequences. Returns a StringView directly into the input
-                // buffer. Switches to the slow path on the first escape hit.
-                //
                 // Two escape forms are recognised, both quote-specific (e.g.
                 // inside a backtick-quoted identifier only backtick escapes
                 // apply; a doubled double-quote is treated as two literals):
                 //   \q  – backslash followed by the opening quote character
                 //   qq  – two consecutive opening quote characters
-                for (; cur_ < end_; ++cur_) {
-                    if (*cur_ == '\\' && cur_ + 1 < end_ && *(cur_ + 1) == quote) {
-                        break; // backslash-escape found, switch to slow path
-                    }
-                    if (*cur_ == quote) {
-                        if (cur_ + 1 < end_ && *(cur_ + 1) == quote) {
-                            break; // doubled-quote escape, switch to slow path
-                        }
-                        const StringView result{start, static_cast<size_t>(cur_ - start)};
-                        ++cur_;
-                        return Token{Token::QuotedIdentifier, result};
-                    }
-                }
-
-                if (cur_ >= end_) {
-                    return Token{Token::Invalid, StringView()};
-                }
-
-                // Slow path: copy content seen so far into scratch_, then
-                // continue scanning and unescaping into it.
-                scratch_.assign(start, cur_);
+                scratch_.clear();
                 for (; cur_ < end_; ++cur_) {
                     if (*cur_ == '\\' && cur_ + 1 < end_ && *(cur_ + 1) == quote) {
                         scratch_ += quote;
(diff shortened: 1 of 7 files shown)


ut/roundtrip_tests.cpp
```diff
@@ -255,6 +259,27 @@ TEST_P(RoundtripCase, TupleTNullableString) {
     EXPECT_TRUE(CompareRecursive(*col, *result_typed));
 }
 
+TEST_P(RoundtripCase, TupleWithQuotedFieldNamesFromServer) {
+    // ClickHouse sends type strings like Tuple(`a.b` Int8) for field names
```
slabko: Is this file really place for this test? I do not really see how it aligns with the there of this test bundle. Maybe `columns_ut.cpp` would be a better place.

slabko (changes requested):
Hi @IyeOnline, can you please move the `TupleWithQuotedFieldNamesFromServer` test to a more appropriate place.
Additionally column and tuple names can have ` in their names. For example:

```
Tuple(`a\`b` UInt8, `a``b` UInt8)
```

slabko:
@IyeOnline did you check the case with inserting data to tables with tuples like ``Tuple(`a.b` Int8)``?


```cpp
// create table test_tuple (
//      id UInt64,
//      value Tuple(`a.b` Int8)
//  )
//  engine = MergeTree
//  order by id

auto block = client.BeginInsert("INSERT INTO test_tuple VALUES");
auto col_id = block.At(0)->AsStrict<ColumnUInt64>();
auto col_tuple = block.At(1)->AsStrict<ColumnTuple>();
auto col_value = col_tuple->At(0)->AsStrict<ColumnInt8>();

col_id->Append(1);
col_id->Append(2);

col_value->Append(5);
col_value->Append(7);
block.RefreshRowCount();

client.SendInsertBlock(block);
client.EndInsert();
```

Gives
```
 DB::Exception: Syntax error (data type): failed at position 8 (.): .b Int8). Expected one of: data type, identifier
```

From what I can see the library does not wrap the names when sending them to the server:

image

IyeOnline:
@slabko I did discover this while trying to send data to a table with a Tuple like that (with the table previously generated from a JSON schema that used such a key) With this fix, it works there. 

I will look into why it doesnt work in the simple standalone case tomorrow.

IyeOnline:
It turns out my code actually didn't _sent_ blocks with column names on the type as that code predates support for that. We just created the table with the named fields.

clickhouse/types/type_parser.cpp
```diff
@@ -260,6 +261,61 @@ TypeParser::Token TypeParser::NextToken() {
...
+                    }
+                    if (*cur_ == quote) {
+                        if (cur_ + 1 < end_ && *(cur_ + 1) == quote) {
+                            break; // doubled-quote escape, switch to slow path
+                        }
+                        const StringView result{start, static_cast<size_t>(cur_ - start)};
+                        ++cur_;
+                        return Token{Token::QuotedIdentifier, result};
+                    }
+                }
```
slabko: Do we need this fast path here? Type name parsing is not most repeated and nor a very long operation.

clickhouse/types/types.cpp
```diff
@@ -473,24 +473,46 @@ LowCardinalityType::LowCardinalityType(TypeRef nested_type) : Type(LowCardinalit
 LowCardinalityType::~LowCardinalityType() {
 }
 
+// Checks if `name` is a valid plain identifier (must not be quoted).
+// The condition for this is a match against `^[a-zA-Z_][0-9a-zA-Z_]*$`
+static bool IsPlainIdentifier(const std::string& name) {
+    if (name.empty()) return false;
+    auto is_alpha_or_under = [](char c) { return std::isalpha(c) || c == '_'; };
+    auto is_alnum_or_under = [](char c) { return std::isalnum(c) || c == '_'; };
```
slabko: `isalpha` and `isalnum` have their own little world of undefined behavior. It is better to avoid them or, at least explicitly convert `c` to `unsigned int`.

slabko (changes requested):
Thank you @IyeOnline for the fixes. Please take a look at my notes.
Additionally, the test for selecting and inserting data are badly missing.

slabko (approved):

PR #511, opened 2026-06-08
Add Bazel e2e test target and fix CityHash wire compatibility

BYVoid:
## Summary

Implements the server-dependent end-to-end tests under Bazel (the follow-up discussed in #510), and fixes a wire-compatibility bug those tests immediately surfaced.

Two independent commits:

1. **CityHash wire-compatibility fix** — a correctness bug in the existing Bazel build, unrelated to the test infra but caught by it.
2. **`//ut:e2e_tests` target + Docker-backed CI** — the actual E2E feature.

## 1. CityHash fix

The merged Bazel build pulls CityHash from the BCR `@cityhash` module, which packages Google's stock CityHash 1.0.2. ClickHouse's wire protocol checksums every compressed block with a **frozen CityHash128 variant** (see `clickhouse/base/compressed.cpp`) that differs from the stock one. Verified directly — same input, different output:

| Source | `CityHash128("The quick brown fox…")` |
|---|---|
| `contrib/cityhash` (what CMake uses) | `69102202d326a2fde4b1346bbee531a1` |
| BCR `@cityhash` (Google f5dc541) | `a7f9a86a2d60c968bf1498f876dbe279` |

So with the BCR module, a server rejects every LZ4/ZSTD block with `DB::Exception: Checksum doesn't match: corrupted data`. Any user of the Bazel-built library with compression enabled hits this. The fix builds the vendored `contrib/cityhash` as a Bazel target (`:cityhash`) and drops the BCR dependency, exactly mirroring the CMake build and guaranteeing byte-identical hashing.

This bug was introduced when the Bazel build switched cityhash to BCR; it went unnoticed precisely because there were no server-backed tests — which is what part 2 adds.

## 2. E2E test target

`//ut:e2e_tests` covers the client round-trip tests that need a live ClickHouse server (`client_ut`, `Column_ut`, `roundtrip_tests`, `low_cardinality_nullable_tests`, `array_of_low_cardinality_tests`, `abnormal_column_names_test`, `readonly_client_test`, `connection_failed_client_test`, `performance_tests`).

- **Tagged `manual` + `external`** — excluded from wildcard expansion, so a bare `bazel test //...` stays hermetic and green without a server, and results are never cached (they depend on live server state).
- **Run it explicitly**, exactly as suggested in #510:
  ```
  bazel run //ut:e2e_tests
  CLICKHOUSE_HOST=my-host bazel test //ut:e2e_tests --test_output=all
  ```
- Connects via the existing `CLICKHOUSE_HOST` / `_PORT` / `_USER` / `_PASSWORD` / `_DB` env vars (default `localhost:9000`).

### CI

The Bazel workflow runs the E2E suite against a live server on **every** matrix platform (3 OS × 3 TLS = 9 jobs), not just Linux. Each OS starts the server the same way the corresponding CMake workflow already does — hosted runners can't all spin up a Linux container identically:

| OS | How the server is started | Mirrors |
|---|---|---|
| Linux | `ci/docker-compose.yml` via `hoverkraft-tech/compose-action` | `linux.yml` |
| macOS | native ClickHouse build downloaded and run as a process (`./clickhouse server`) | `macos.yml` |
| Windows | the Linux container under WSL2 + `podman-compose` | `windows_msvc.yml` |

All three expose the server on `localhost:9000`, which is the E2E suite's default target, so the test invocation (`bazel test //ut:e2e_tests`) is identical across platforms. `//ut:e2e_tests` is also built explicitly on every platform first — the `manual` tag means `bazel build //...` skips it — giving compile/link coverage everywhere even though the server run is what exercises it.

### Notes / scope

- **No runfiles needed.** These tests are pure env-driven network clients — none read data files — so the runfiles plumbing I anticipated in #510 turned out unnecessary. (If a future test needs fixtures, that's where runfiles would come in.)
- `ssl_ut.cpp` is intentionally **not** included: it connects to an external secure endpoint (`play.clickhouse.com:9440`) with OS-specific CA paths, which the local docker server doesn't provide. Covering it would mean configuring a TLS-enabled server in CI; left for a separate change.

## Test plan

- [x] `bazel test //...` runs only `//ut:unit_tests` (e2e correctly excluded by `manual`), green.
- [x] `bazel build //ut:e2e_tests` succeeds for `tls=boringssl|openssl|no`.
- [x] `bazel test //ut:e2e_tests` against a local docker ClickHouse 25.12: **665 passed**, 50 skipped. With the BCR cityhash (before the fix) the compression round-trips failed with checksum-mismatch; after the fix they pass.
- [x] `bazel run //ut:e2e_tests -- --gtest_filter=...` passes gtest args through.


slabko (approved):
Awesome!
Clean an simple

slabko:
@BYVoid,

The version of CityHash used by ClickHouse is 1.0.1 (this commit gives the expected hash https://github.com/google/cityhash/commit/8eded14d8e7cabfcdb10d4be35d521683edc0407)

Do you think it would be a good idea to include CityHash 1.0.1 to BCR?

Another option is to completely hide CityHash within the library by renaming its symbols. This would allow users to use any version of CityHash without encountering symbol conflicts. This may be the better approach, since we are likely to keep this version indefinitely and it is used only internally.

PR #513, opened 2026-06-09
Rename Bazel module to clickhouse-cpp-client

BYVoid:
## Summary

Renames the Bazel module from `clickhouse_cpp` to **`clickhouse-cpp-client`**.

The name was discussed with folks from the Bazel community, and the conclusion was that `clickhouse-cpp-client` better reflects what the package actually is: a **client** library for talking to ClickHouse. A bare `clickhouse-cpp` is easy to misread as the C++ ClickHouse *server* (or as bundling server code), whereas the `-client` suffix makes the scope unambiguous. This is also the name the module is already published under on the [Bazel Central Registry](https://registry.bazel.build/), so aligning the in-repo `MODULE.bazel` removes the divergence between the source and the registry.

Until now the BCR entry carried an overlay that renamed the module, and the build-setting flag consumers used (`--@clickhouse_cpp//:tls=...`) didn't match the published name. After this change the repo and BCR agree, and future releases need no overlay rename.

## Changes

| File | Change |
|---|---|
| `MODULE.bazel` | `module(name = "clickhouse-cpp-client", …)` |
| `BUILD.bazel` | doc comments now reference `@clickhouse-cpp-client//:tls` / `@clickhouse-cpp-client//:clickhouse` |
| `.github/workflows/bazel.yml` | the four `--@clickhouse-cpp-client//:tls=${{ matrix.tls }}` invocations |

No `MODULE.bazel.lock` change (the lock tracks dependencies, not the root module name); the dependency graph is untouched.

## Flag usage after the rename

```
bazel build //:clickhouse                                          # BoringSSL (default)
bazel build //:clickhouse --@clickhouse-cpp-client//:tls=openssl   # OpenSSL from BCR
bazel build //:clickhouse --@clickhouse-cpp-client//:tls=no        # TLS disabled
```

## Test plan

- [x] `bazel build //... --@clickhouse-cpp-client//:tls=openssl` succeeds.
- [x] `bazel test //ut:unit_tests --@clickhouse-cpp-client//:tls=no` passes.
- [x] `bazel build //ut:e2e_tests --@clickhouse-cpp-client//:tls=boringssl` succeeds.
- [x] No stale `clickhouse_cpp` references remain in any Bazel/CI file.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -40,16 +40,16 @@ jobs:
           repository-cache: true
 
       - name: Build project
-        run: bazel build //... --@clickhouse_cpp//:tls=${{ matrix.tls }}
+        run: bazel build //... --@clickhouse-cpp-client//:tls=${{ matrix.tls }}
 
       - name: Run unit tests
-        run: bazel test //ut:unit_tests --@clickhouse_cpp//:tls=${{ matrix.tls }} --test_output=errors
+        run: bazel test //ut:unit_tests --@clickhouse-cpp-client//:tls=${{ matrix.tls }} --test_output=errors
 
       # //ut:e2e_tests is tagged `manual`, so `bazel build //...` above
       # skips it; build it explicitly so the run step below doesn't pay
       # for compilation under the running-server timeout.
       - name: Build e2e tests
-        run: bazel build //ut:e2e_tests --@clickhouse_cpp//:tls=${{ matrix.tls }}
+        run: bazel build //ut:e2e_tests --@clickhouse-cpp-client//:tls=${{ matrix.tls }}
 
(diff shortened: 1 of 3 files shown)


slabko (approved):

slabko:
@BYVoid, 2.6.2 is now live. I think it would also be a good idea to mention in the release notes that the library is available in the BCR under the name `clickhouse-cpp-client`. I haven't added that yet because 2.6.2 is not available there yet. As soon as it is, I will update the release notes and README.md.

BYVoid:
https://github.com/bazelbuild/bazel-central-registry/pull/9202 Can you please comment there that you plan to become the maintainer of the package?

slabko:
@BYVoid,

Wtote it at first, but I noticed that you have already changed and merged it. Do I still need to do that?

PR #516, opened 2026-06-10
client: clear Inserting state before connect in ResetConnection

iliaal:
ResetConnection() connects first, then clears the inserting state.

When the reconnect throws (max_connections, port exhaustion, DNS blip, etc.), the state stays Inserting. The destructor then calls EndInsert() on the old dirty socket and commits whatever the caller had in flight.

Move state_ = State::Idle to the first line, before any connect. A failed reconnect now safely prevents the implicit EndInsert.

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -606,6 +606,13 @@ void Client::Impl::Ping() {
 }
 
 void Client::Impl::ResetConnection() {
+    // Clear the insert state BEFORE attempting to reconnect. If connect()
+    // throws (server at max_connections, ephemeral-port exhaustion, DNS blip,
+    // etc. — all cases where the original dirty socket may still be healthy),
+    // a still-Inserting state would make ~Impl's automatic EndInsert() fire
+    // on the old dirty wire and commit a partially-streamed insert.
+    state_ = State::Idle;
+
     InitializeStreams(socket_factory_->connect(options_, current_endpoint_.value()));
     state_ = State::Idle;
 


slabko:
Hi, @iliaal,

The line above `state_ = state::Idle` provides the strong exception guarantee, meaning that if anything in it throws, the client's state remains unchanged. Changing the state before this line completes is not a good idea because it breaks this exception-safety guarantee. If you encounter a temporary network issue while attempting to reconnect, the client will retain the old socket (if one existed before), along with its current state.

When this happens during a reconnect (in `RetryGuard`), the same rule applies: the old broken socket will remain assigned to the client until the next attempt to `ResetConnection`, so it makes sense to preserve the client's state as well.

Could you please explain your use case? I can't really see a situation in which moving this line earlier would be beneficial.

iliaal:
Use case: a user-initiated `Client::ResetConnection()` called mid-insert (after `BeginInsert`/`SendInsertBlock`) to drop a half-streamed insert and rebuild the connection.

You're right that this weakens the strong exception guarantee of `ResetConnection` in isolation, and my "the old socket may still be healthy" comment was misleading. The real problem is the destructor. `~Impl` calls `EndInsert()` whenever `state_ == Inserting`, and a failed `connect()` never reaches `InitializeStreams`, so that `EndInsert` runs on the *old* socket and commits a partial insert (or throws into the swallowed `catch(...)`). A user-initiated reset means "drop the in-flight insert," so auto-committing it is wrong.

`RetryGuard`'s internal reconnects always run with `state_ == Idle`. All three call sites (`BeginExecuteQuery`, `Insert`, `BeginInsert`) invoke `RetryGuard` before they leave `Idle`, so this line is a no-op on internal reconnects and only affects the user-facing reset.

If you'd rather keep `Impl::ResetConnection` strictly exception-safe, I can move the insert-drop into the public `Client::ResetConnection` instead: internal retries keep the strong guarantee, only the user-initiated reset drops the insert.

slabko:
@iliaal,

I see the problem now. That makes sense. However, the issue is that inserts in ClickHouse are not transactional.

If you insert a small dataset, it may appear as though resetting the connection without calling `EndInsert` provides a way to roll back the insert. However, if you continue sending data or the inserted block is large enough, ClickHouse will eventually flush the data, even if `EndInsert` is never called.

As a result, you can never rely on resetting the connection and omitting `EndInsert` as a mechanism for rolling back inserts.

iliaal:
Agreed: inserts aren't transactional, and I'm not treating skip-EndInsert as a rollback. Whatever the server already flushed stays flushed; that's its at-least-once behavior and outside the client's control. The fix doesn't depend on that.

The defect is narrower. After a user calls `ResetConnection` mid-insert, if `connect()` to the new socket throws, `InitializeStreams` never runs, so `output_` still points at the old wire. With `state_` still `Inserting`, `~Impl` then calls `EndInsert()` on that old socket. Two bad outcomes:

1. The old socket is still alive (server hit max_connections on the second connection, DNS blip on a rotated endpoint, and so on). `EndInsert` sends the empty end-of-data marker and finalizes the insert the user just asked to drop, including the last buffered block. That's the client completing the handshake, not the server's flush behavior.
2. The old socket is dead. `EndInsert` does blocking I/O (`SendData` plus the `ProcessPacket` EOS loop) inside a destructor, then throws into the swallowed `catch(...)`.

So the guarantee is "don't complete the insert handshake for an insert the user abandoned," not "atomic rollback."

If you'd rather keep `Impl::ResetConnection` strictly exception-safe, my earlier offer still stands: move the state clear into the public `Client::ResetConnection` so `RetryGuard`'s internal retries keep the strong exception guarantee and only the user-initiated reset drops the insert. Happy to push that; your call on placement.

slabko:
@iliaal, I do not understand the "don't complete the insert handshake for an insert the user abandoned" guarantee. We do not provide such a guarantee. This is an unfortunate property of many objects that cannot simply be abandoned and must instead be wound down somehow. `std::fstream` exhibits similar behavior and attempts to close a file on destruction (flushing it beforehand).

I also have difficulty understanding your use case. If any of the issues you listed occur, then the guarantee you are asking for would certainly be nice to have, but it is not critical. Exception guarantees, on the other hand, are very important.

iliaal:
Fair, and the `std::fstream` parallel lands. I'll drop the "abandon the insert" framing: the library never promised it, and the strong exception guarantee on `ResetConnection` is the more important property. Not worth trading it for this.

The one residual I'd note, separate from insert semantics: after a failed `connect()`, `output_` still points at the old socket, so the destructor's `EndInsert()` runs network I/O on a wire that's already being torn down. That's a teardown-robustness question, not a rollback one. If you don't see it as worth the added complexity, I'm happy to close this.

slabko:
I just don't really understand what your actual use case is. Is this a real problem you're encountering, or is it more of a theoretical concern?

So far, it seems to me that we're trying to determine which inconvenience is more significant: exception guarantees or clearing the connection state after an error.

iliaal:
Real, not theoretical. It's a live recovery path in the php_clickhouse extension, the native binding I maintain on top of this client.

The binding exposes streaming inserts (`writeStart` / `write` / `writeEnd`) backed by `BeginInsert` / `SendInsertBlock` / `EndInsert`. When a PHP object holding a half-streamed insert is destroyed without `writeEnd` (script bailout, exception unwind, `unset()` mid-insert), the destructor has to decide what to do with the in-flight insert. Committing it silently with `EndInsert` would turn an aborted script into a partial commit, so once any block has been sent the binding calls `ResetConnection()` to discard it:

```cpp
// clickhouse_free_obj()
if (obj->client && obj->has_insert_block) {
    if (obj->insert_blocks_sent)
        try { obj->client->ResetConnection(); } catch (...) {}  // discard partial insert
    else
        try { obj->client->EndInsert();      } catch (...) {}  // clean session, close empty
}
if (obj->client) { delete obj->client; ... }                   // ~Impl runs here
```

When `connect()` succeeds, `state_` goes `Idle` and the following `delete` is clean. When `connect()` throws (server at max_connections, ephemeral-port exhaustion at teardown under load, a DNS/failover blip), the throw is caught but `state_` stays `Inserting`. The next line deletes the client, `~Impl` sees `Inserting`, and `EndInsert()` runs on the old socket, committing the exact partial insert the binding just tried to discard.

The crux: a consumer has no way to abandon an in-flight insert across a failed reconnect. `EndInsert` commits; the only non-committing way to clear the `Inserting` state is a *successful* `ResetConnection`. When the reconnect fails, the destructor defaults to commit, and the binding can't suppress it from outside; there's no public API to clear the state.

That's narrower than the "rollback guarantee" I muddied this with earlier. Two ways to close it without touching the exception guarantee you care about: move the state clear into the public `Client::ResetConnection` (internal `RetryGuard` retries keep the strong guarantee, only the user-facing reset clears state up front), or add an explicit abort call so a consumer can say "drop this insert" without going through `EndInsert`. Either works for me; if neither appeals, I'll close.

slabko:
```cpp
// clickhouse_free_obj()
if (obj->client && obj->has_insert_block) {
    if (obj->insert_blocks_sent)
        try { obj->client->ResetConnection(); } catch (...) {}  // (1) discard partial insert
    else
        try { obj->client->EndInsert(); } catch (...) {}        // (2) clean session, close empty
}
if (obj->client) { delete obj->client; ... }
```

What really happens in (1) is that you don't just close the connection, you actually create another connection and perform a handshake just to close it immediately afterward. All of this is done for the questionable goal of avoiding partial writes, but partial writes may still occur anyway, because that's how ClickHouse works.

Also note that (2) does nothing useful, because the subsequent `delete obj->client` effectively does the same thing.

In reality, the only thing you need to do is call the destructor:

```cpp
// clickhouse_free_obj()
if (obj->client) { delete obj->client; ... }
```

Trying to avoid partial writes and the side effects of closing the connection provides very little benefit and is likely to be a source of confusion for many, because partial writes can still occur in most cases. For this reason, the library is designed not to attempt to prevent them. Instead, it tries to make this behavior as explicit as possible.

iliaal:
I see, thanks for the explanation that makes sense, I'll make the changes in the extension.

PR #517, opened 2026-06-10
client: pass full Query to SendQuery in BeginInsert (preserve query_id)

iliaal:
BeginInsert(Query query) was calling SendQuery(query.GetText()).

The implicit Query(string) constructor uses an empty default query id, so any query_id the caller supplied was silently dropped. The INSERT would land in system.query_log with an auto-generated id.

Pass the full Query object (the one-shot Insert path already did this).

This was originally carried as a local patch and is exercised by real usage that cares about stable query ids for auditing.

### DIFF
--- a/clickhouse/client.cpp
+++ b/clickhouse/client.cpp
@@ -546,7 +546,7 @@ Block Client::Impl::BeginInsert(Query query) {
         return true;
     });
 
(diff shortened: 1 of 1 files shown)


slabko (approved):
@iliaal, this is indeed a bug, thank you for caching it!

PR #518, opened 2026-06-10
client: add BeginInsert(const Query&) overload

iliaal:
Add Block BeginInsert(const Query& query).

The Impl already supported a fully-configured Query (settings, params, query_id, OnData, etc.). The public surface only exposed the string forms.

This lets callers do proper streaming inserts with the full Query feature set (symmetric to the Query forms already available for Select/Execute).

### DIFF
--- a/clickhouse/client.h
+++ b/clickhouse/client.h
@@ -306,9 +306,8 @@ public:
     void Insert(const std::string& table_name, const std::string& query_id, const Block& block);
 
     /// Start an \p INSERT statement, insert batches of data, then finish the insert.
-    /// A bare string converts to a Query implicitly; use BeginInsert(const Query&)
-    /// to pass per-insert settings, params, and query_id. Event callbacks on the
-    /// Query are rejected (throws ValidationError): BeginInsert owns the data path.
+    /// Queries with event callbacks are not allowed.If the query has any event callbacks set, this
+    /// call throws ValidationError.
     Block BeginInsert(const Query& query);
     Block BeginInsert(const std::string& query, const std::string& query_id);
 
--- a/clickhouse/client.h
+++ b/clickhouse/client.h
@@ -302,8 +302,10 @@ public:
     void Insert(const std::string& table_name, const std::string& query_id, const Block& block);
 
     /// Start an \p INSERT statement, insert batches of data, then finish the insert.
-    /// Use BeginInsert(const Query&) for settings, query_id, callbacks etc.
+    /// A bare string converts to a Query implicitly; use BeginInsert(const Query&)
+    /// for settings, params, callbacks etc.
     Block BeginInsert(const Query& query);
+    Block BeginInsert(const std::string& query, const std::string& query_id);
 
     /// Insert data using a \p block returned by \p BeginInsert.
     void SendInsertBlock(const Block& block);
--- a/clickhouse/client.h
+++ b/clickhouse/client.h
@@ -303,7 +303,8 @@ public:
 
     /// Start an \p INSERT statement, insert batches of data, then finish the insert.
     /// A bare string converts to a Query implicitly; use BeginInsert(const Query&)
-    /// for settings, params, callbacks etc.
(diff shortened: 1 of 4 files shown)


clickhouse/client.h
```diff
@@ -302,8 +302,8 @@ class Client {
     void Insert(const std::string& table_name, const std::string& query_id, const Block& block);
 
     /// Start an \p INSERT statement, insert batches of data, then finish the insert.
-    Block BeginInsert(const std::string& query);
-    Block BeginInsert(const std::string& query, const std::string& query_id);
+    /// Use BeginInsert(const Query&) for settings, query_id, callbacks etc.
+    Block BeginInsert(const Query& query);
```
slabko: Why should we take away a part of existing API on which people probably relay? Can `BeginInsert(const Query& query)` be an addition to the exiting functions?

slabko:
The main reason `BeginInsert` does not take a `Query` as a parameter is that `Query` instances are much more than just a query string and a query ID. They also contain a number of event handlers. These event handlers add significant complexity to the API, and it is not entirely clear how they should be handled in this context.

To avoid this issue, the decision was made not to expose the `Query` object in such cases. I still stand by that decision because exposing it would create an unforeseen set of use cases that people would come to rely on, making them difficult to support in the future. I am not comfortable extending the API surface area that far just yet.

If you can explain why you need it, we might try to find another solution to it.

iliaal:
The driver is per-insert settings, not the handlers.

In my binding every streaming insert attaches `Query::SetSetting(...)` (async_insert, format settings, plus per-call overrides) and then calls `BeginInsert`. With only the string forms there's no way to scope settings to a single insert. I'd have to mutate the connection's global settings, which then leak into the next query on the same handle. The `Query` overload is the only way to pass settings (and bound parameters) without that shared state.

On the handlers: `BeginInsert` already neutralizes them. `Impl::BeginInsert` overwrites the caller's `OnData` with its own block-capture lambda, and the `EnsureNull` guard resets `events_` to null the moment `BeginInsert` returns. So no handler on the passed `Query` survives into the `SendInsertBlock`/`EndInsert` phase. The data path runs without callbacks regardless of what the `Query` carries. The only window anything fires is the initial handshake inside `BeginInsert`, and even there `OnData` is replaced.

If exposing the full `Query` still feels too broad, an overload taking just `Settings` (and `query_id`) covers my case without touching the callback surface. Happy to go either way.

slabko:
> The data path runs without callbacks regardless of what the Query carries. The only window anything fires is the initial handshake inside BeginInsert, and even there OnData is replaced.

I'm not sure that's entirely true. The first thing that comes to mind is a server exception.

If we go down this route, could we throw an exception stating that `Query` callbacks are not supported for `BeginInsert` whenever any `Query` callback is set?

iliaal:
You're right, my "without callbacks regardless" was wrong. `BeginInsert` only overwrites `OnData`. `EnsureNull` points `events_` at the passed `Query` for the whole call, so during the handshake packet loop the server can still drive `OnServerException`, `OnProgress`, `OnProfile`, `OnServerLog`, `OnProfileEvents`, and `OnDataCancelable` into whatever the caller set. The exception path is the clearest case.

Pushed the guard you suggested. Before `BeginInsert` installs its own `OnData`, it now rejects any `Query` that carries a handler:

```cpp
if (query.HasEventCallbacks())
    throw ValidationError("Query callbacks are not supported in BeginInsert");
```

The callbacks are private `std::function` members, so I added a small predicate on `Query`:

```cpp
inline bool HasEventCallbacks() const {
    return exception_cb_ || progress_cb_ || select_cb_ || select_cancelable_cb_
        || select_server_log_cb_ || profile_events_callback_cb_ || profile_callback_cb_;
}
```

That keeps the overload scoped to per-insert settings, params, and query_id, with the unsupported path a hard error instead of a silent surprise. Added a unit test covering both the rejection and a settings+query_id streaming insert; passes against a local server.

slabko (approved):

PR #522, opened 2026-06-17
Use BCR cityhash module

BYVoid:
## Summary
- add the BCR `cityhash@1.0.2.bcr.1` module dependency
- replace the Bazel-local vendored CityHash target with `@cityhash//:cityhash`
- remove the stale note that BCR cityhash is hash-incompatible; `1.0.2.bcr.1` now packages the frozen CityHash 1.0.2 source with the `ssize_t` compatibility patch

## Why this can replace the Bazel part of #515
#515 fixes the Bazel-local `:cityhash` target under the assumption that Bazel keeps compiling `contrib/cityhash`. In that setup, the vendored `city.cc` included a generic `config.h`, so #515 renamed the checked-in header to `city_config.h`, prefixed its macros, and updated the local Bazel `srcs` list.

This PR removes that Bazel-local target entirely. Bazel no longer compiles `contrib/cityhash`; it consumes CityHash through the BCR module instead. That makes the Bazel-specific part of #515 unnecessary: there is no local `:cityhash` rule, no local `city_config.h` entry to maintain, and no local Windows `/DWIN32` workaround for CityHash. The CMake/vendor-side cleanup from #515 can still be useful for the checked-in contrib copy, but Bazel no longer needs to own or patch that copy.

In short: #515 makes the vendored Bazel CityHash target safer; this PR stops vendoring CityHash in Bazel. The latter is a smaller maintenance surface for Bazel.

## Why `cityhash@1.0.2.bcr.1` matches the local contrib version
The local `contrib/cityhash` copy is based on upstream `google/cityhash@bc38ef45ddbbe640e48db7b8ef65e80ea7f71298`, i.e. CityHash 1.0.2. Later local changes were compatibility/build-system edits, not intentional hash algorithm changes.

The BCR `cityhash@1.0.2.bcr.1` module points to that same upstream commit and applies the same relevant compatibility fix for `ssize_t` (`remove_ssize_t.patch`). That patch matches the local `c02157e`-style fix and preserves the CityHash64/CityHash128 behavior.

I also verified this locally by comparing outputs from patched BCR CityHash and the checked-in contrib CityHash for multiple input lengths across:
- `CityHash64`
- `CityHash64WithSeed`
- `CityHash128`

The outputs were identical, and the server-dependent e2e tests with compressed roundtrips passed against a local ClickHouse server.

## Testing
- `bazel --ignore_all_rc_files test @cityhash//:all --registry=https://bcr.bazel.build`
- `bazel --ignore_all_rc_files test //... --registry=https://bcr.bazel.build`
- `bazel --ignore_all_rc_files test //ut:e2e_tests --test_output=errors --cache_test_results=no --registry=https://bcr.bazel.build`

Also tested earlier with `--//:tls=no` and `--//:tls=openssl` variants while validating the cityhash swap.


slabko:
Thank you for following up on CityHash, and for the careful verification work in this PR.

After thinking it over, I've concluded that a better direction is to fully hide our CityHash dependency rather than source it externally: rename its symbols and wrap them in a dedicated namespace, strip out whatever parts of CityHash the project doesn't use, and move the trimmed copy out of `contrib/` into the project tree proper.

A few reasons this fits us better than consuming CityHash through BCR:

- It keeps the version pinned (as it effectively is today), but without relying on an external module that a consumer could override.
- It gives us the freedom to evolve the code and ClickHouse's checksum algorithm on our own terms if and when that becomes necessary.
- Most importantly, it lets downstream users link any version of CityHash they like without colliding with ClickHouse's internal copy, which is the actual goal of avoiding the symbol clash, and something the shared/BCR symbol still wouldn't solve.

So I'd suggest we leave everything as is for now and hold off on this change until I've moved the pinned CityHash into the project. Really appreciate you digging into the version compatibility.

PR #523, opened 2026-06-25
Add Abseil-free 128-bit integer implementation

slabko:
Introduce an abstraction layer for wide (128-bit) integers so the library can be built without Google Abseil. The new CH_USE_ABSEIL_FOR_BIGNUM CMake option (ON by default) selects between Abseil's int128/uint128 and a self-contained fallback implementation.

Library:
- Add clickhouse/types/bignum.{h,cpp} defining Int128/UInt128 (Abseil-backed or fallback structs) plus a static Bignum helper class providing a uniform API (Make/Low64/High64, string conversions). The fallback type provides comparisons, numeric_limits, ostream output, and static_asserts guarding the 16-byte little-endian layout required for raw wire serialization.
- Add clickhouse/base/bignum_string.h, a low-level limb-array <-> decimal string converter supporting 128/256/512-bit widths.
- Rewrite ColumnDecimal::Append(string) to validate precision/scale up front and normalize the value before parsing via Bignum::StringToInt128, replacing the hand-rolled overflow-checked accumulation.
- Route Int128 access in ColumnDateTime64 and ColumnDecimal through Bignum helpers; drop the Int128/UInt128 aliases from numeric.h and types.h in favor of bignum.h.

Tests/CI:
- Add bignum unit tests (round-trip, samples, string conversion) and update client/columns/itemview tests to use Bignum::Int128ToString.
- Build the test matrix with CH_USE_ABSEIL_FOR_BIGNUM both ON and OFF.

The longer-term goal is to remove Abseil from the library entirely and let users plug in their own bignum library of choice.


slabko:
@BYVoid can you please check that I didn't miss anything in the last commit (removing Abseil from Bazel build)?

BYVoid:
LGTM

chernser (approved):

Issue #477, opened 2026-04-10
Add `GetString()` to `ColumnDecimal`


iskakaushik:
pg_clickhouse reads Decimal columns by manually converting the internal Int128 representation to a formatted decimal string (handling sign, scale, zero-padding, and decimal point placement). This is ~50 lines of custom code. A GetString() method on ColumnDecimal that returns the properly formatted value would let us replace this with a single call.

theory:
See #451 for a working implementation. Could be optimized later without changing the interface.

PR #526, opened 2026-06-27
Add StringAt to ColumnDecimal

slabko:
Adds `ColumnDecimal::StringAt(size_t)`, which returns the human-readable decimal string representation of the value at index `i`, with the decimal point inserted according to the column's scale.

Solves https://github.com/ClickHouse/clickhouse-cpp/issues/477

### DIFF
--- a/ut/columns_ut.cpp
+++ b/ut/columns_ut.cpp
@@ -124,6 +124,7 @@ TEST(ColumnsCase, DecimalStringValueMapping) {
     };
     
     std::vector<TestSample> samples = {
+        {18, 0, "0.0", Int128(0)},
         {18, 0, "0.123", Int128(0)},
         {18, 0, "123", Int128(123)},
         {18, 3, "0.123", Int128(123)},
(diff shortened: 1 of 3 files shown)


ut/columns_ut.cpp
```diff
@@ -144,6 +145,45 @@ TEST(ColumnsCase, DecimalStringValueMapping) {
...
+TEST(ColumnsCase, DecimalStringAt) {
+    struct TestSample {
+        size_t precision;
+        size_t scale;
+        std::string str;
+        std::string expect;
+    };
+
+    std::vector<TestSample> samples = {
+        {18, 0, "0", "0"},
```
kavirajk, line 157: any invalid string case? e.g: {18, 0, "invalid"} and expect an exception?

kavirajk (approved):

ut/columns_ut.cpp
slabko: This test checks that numeric values are correctly converted to strings. So input always should be good. There is a separate test that checks the parsing itself.  This one for bad strings:
https://github.com/ClickHouse/clickhouse-cpp/blob/master/ut/columns_ut.cpp#L70

PR #527, opened 2026-06-27
Add native __int128 casts for bignum types

slabko:
Expose explicit conversions from Int128 and UInt128 to compiler-native 128-bit integers when `__int128` is available. Add unit coverage for signed and unsigned casts across zero, small values, large values, and boundary limits.

### DIFF
--- a/clickhouse/types/bignum.h
+++ b/clickhouse/types/bignum.h
@@ -4,6 +4,13 @@
 #include <string>
 #include <string_view>
 
+
+ #if defined(__SIZEOF_INT128__)                                                                       
+ #define CH_CPP_HAS_INT128 1                                                                  
+ #else                                                                                                
+ #define CH_CPP_HAS_INT128 0                                                                  
+ #endif                                                                                               
+
 /**
  * This file contains declarations and definitions of the types and API used for wide
  * (128-, 256- bit) integers. It first declares the Int128 and UInt128, implementation of which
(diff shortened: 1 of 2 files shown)


clickhouse/types/bignum.h
```diff
@@ -66,6 +73,12 @@ struct UInt128 {
     bool operator<=(const UInt128& other) const { return !(other < *this); }
     bool operator>=(const UInt128& other) const { return !(*this < other); }
 
+#if CH_CPP_HAS_INT128
+    explicit operator unsigned __int128() const {
+        return ((unsigned __int128)limbs[1] << 64 | limbs[0]);
```
kavirajk, line 78: not an expert. why not cpp style `static_cast` here? instead of C-style cast? Curious.

clickhouse/types/bignum.h
```diff
@@ -104,6 +117,12 @@ struct Int128 {
     bool operator<=(const Int128& other) const { return !(other < *this); }
     bool operator>=(const Int128& other) const { return !(*this < other); }
 
+#if CH_CPP_HAS_INT128
+    explicit operator __int128() const {
+        return (__int128)((unsigned __int128)limbs[1] << 64 | limbs[0]);
```
kavirajk, line 122: same here.

kavirajk (approved):
LGTM. one minor clarification.

clickhouse/types/bignum.h
slabko: No specific reason, normally I stick to `static_cast`, however I sometimes bypass it, but only when I work with numerics and it gets too verbose. Otherwise it would be:

```
return static_cast<__int128>(static_cast<unsigned __int128>(limbs[1]) << 64 | limbs[0]);
```

PR #530, opened 2026-06-30
Support writing generic LowCardinality(T)

IyeOnline:
Previously, clickhouse-cpp only supported writing LowCardinality(String).
With this change, support is expanded to other fundamental types as well.


slabko:
@IyeOnline

I am a bit puzzled by this. Non-string `LowCardinality` types aren’t very popular, and in most cases, using plain numeric types is faster. They avoid unnecessary lookups at the cost of using a bit more storage, though not always even that. ClickHouse even disables them by default and requires enabling `allow_suspicious_low_cardinality_types`.

Where and how do you use them? I’m wondering what the use case is here.

IyeOnline:
@slabko There actually is no immediate use.

I was adding LowCardinality support to allow sending data to LowCardinality columns (they are strings in the concrete case) and noticed that it wouldnt work for non-string columns.

Since our sink is user-controlled, I would like to support this - even if its an evidently bad idea for most columns and most likely would be rejected by a default-configured ClickHouse server anyways.

mavam:
We have users that use these types, even though they might not be recommended or there are reasons why numeric types would be preferred. We currently cannot work with these Clickhouse deployments, and it's important to us that we can serve all users. These are effectively supported type and many people use them in prod.

slabko:
@IyeOnline,

Just wondering, how you did you make `DateTime64` an `Enum` work with `LowCardinality ` in ClickHouse?

slabko (approved):

PR #533, opened 2026-07-02
Vendor CityHash as an internal part of the library

slabko:
Vendor CityHash as an internal part of the library

ClickHouse's wire protocol checksums compressed blocks with a specific
frozen CityHash128 variant. It is not interchangeable with upstream or
system CityHash: a different implementation produces different hashes and
the server rejects blocks with "Checksum doesn't match: corrupted data".

Treating it as a swappable contrib/ dependency was therefore misleading
and unsafe. Move it out of contrib/ and into the library proper:

- Move contrib/cityhash/cityhash/ -> clickhouse/cityhash/.
- Wrap the code in a clickhouse::cityhash namespace to prevent symbol
  clashes when consumers link their own CityHash; update call sites in
  compressed.cpp, lowcardinality.cpp and types.cpp accordingly.
- Drop the WITH_SYSTEM_CITYHASH option, which could silently break wire
  compatibility, along with the Findcityhash machinery it required.
- Compile CityHash directly into the clickhouse target (Bazel and CMake)
  instead of as a separate library.
- Remove the cityhash Bazel target and move it to clikchouse target itself.


slabko:
@BYVoid, if you have a moment, can check if this change is ok https://github.com/ClickHouse/clickhouse-cpp/pull/533/changes/2f1783a248b38361ee8a082c77fc0f49e9d4e086

BYVoid:
LGTM

joe-clickhouse (approved):
LGTM

PR #541, opened 2026-07-29
Typed column access with `As<>` / `AsStrict<>`

slabko:
## TL;DR

`Column::As<T>()` and `Column::AsStrict<T>()` are now smarter for the *typed* column
templates (`ColumnArrayT`, `ColumnTupleT`, `ColumnMapT`, `ColumnNullableT`,
`ColumnLowCardinalityT`). When the requested `T` is one of these and the column is not
already exactly `T`, they **wrap** the column into a strongly-typed, storage-sharing view
instead of failing.

In practice this means you can take a column straight out of a `Select` result and ask for
the convenient typed interface directly:

```cpp
// block[0] is a base ColumnLowCardinality coming from the server.
if (auto lc = block[0]->As<ColumnLowCardinalityT<ColumnString>>()) {
    for (size_t i = 0; i < lc->Size(); ++i)
        std::cout << (*lc)[i] << "\n";   // typed operator[] -> std::string_view
}
```

No manual reconstruction, no exact-type juggling.

---

## Background: base columns vs. typed columns

ClickHouse composite types are represented by two layers in this library:

| Base column           | Typed wrapper                              | Adds                                            |
|-----------------------|--------------------------------------------|-------------------------------------------------|
| `ColumnArray`         | `ColumnArrayT<Nested>`                      | typed `At()`, `operator[]`, `Append(container)` |
| `ColumnNullable`      | `ColumnNullableT<Nested>`                   | `std::optional`-based `At()` / `operator[]`     |
| `ColumnTuple`         | `ColumnTupleT<Cols...>`                     | `std::tuple`-based `At()` / `operator[]`        |
| `ColumnMap`           | `ColumnMapT<Key, Value>`                    | key/value `At()`, iteration                     |
| `ColumnLowCardinality`| `ColumnLowCardinalityT<Dict>`               | typed `At()`, `operator[]`, `AppendMany()`      |

When you `Select`, the server-side columns are created by the factory as the **base**
types (`ColumnArray`, `ColumnNullable`, `ColumnTuple`, `ColumnMap`, `ColumnLowCardinality`).
The `…T<>` templates are the ergonomic, type-safe wrappers around them.

Previously, to get a typed view you needed the column to already *be* that exact type, or
you had to build the typed column yourself. Now `As<>`/`AsStrict<>` bridge the gap
automatically.

---

## Behavior

A type is **wrappable** if it exposes a static `Wrap` method. That is exactly the five
templates above (`ColumnArrayT`, `ColumnTupleT`, `ColumnMapT`, `ColumnNullableT`,
`ColumnLowCardinalityT`). Everything else (`ColumnString`, `ColumnUInt64`, …) is
non-wrappable and behaves as before.

| Call                    | Wrappable `T`                                                        | Non-wrappable `T`                       |
|-------------------------|----------------------------------------------------------------------|-----------------------------------------|
| `As<T>()` (non-const)   | exact cast first; otherwise wrap. Returns `nullptr` on mismatch.     | plain cast; `nullptr` on mismatch.      |
| `AsStrict<T>()`         | exact cast first; otherwise wrap. Throws `ValidationError` on failure.| plain cast; throws on failure.          |
| `As<T>() const`         | **exact cast only — no wrapping**; `nullptr` on mismatch.            | plain cast; `nullptr` on mismatch.      |

Notes:

- The exact-cast-first step means that if the column already *is* `T`, you get the very same
  object back (identity preserved, no new allocation).
- `Block::operator[]` returns a non-const `ColumnRef` by value, so `block[i]->As<…>()`
  always uses the wrapping (non-const) overload — wrapping works out of the box for `Select`
  results.

### Wrapping is a storage-sharing view

Wrapping does **not** copy or move the underlying data. The typed wrapper shares the same
storage (dictionary, offsets, index, nested columns) with the source via `shared_ptr`.
Mutations through one are visible through the other, and the source stays fully valid.

```cpp
auto base    = block[0]->As<ColumnArray>();                 // base view
auto typed   = block[0]->As<ColumnArrayT<ColumnUInt64>>();  // typed, shares the same storage
// reading through `typed` reflects exactly what `base` holds
```

---

### LowCardinality

```cpp
client.Select("SELECT lc_col FROM my_table", [](const Block& block) {
    // Server sends a base ColumnLowCardinality; wrap it into the typed view.
    auto lc = block[0]->As<ColumnLowCardinalityT<ColumnString>>();
    if (!lc) return;
    for (size_t i = 0; i < lc->Size(); ++i)
        std::cout << lc->At(i) << "\n";   // std::string_view
});
```

### Array

```cpp
// Array(UInt64)
auto arr = block[0]->As<ColumnArrayT<ColumnUInt64>>();
for (size_t row = 0; row < arr->Size(); ++row) {
    auto values = arr->At(row);                 // ArrayValueView
    for (size_t j = 0; j < values.size(); ++j)
        std::cout << values[j] << " ";          // uint64_t
    std::cout << "\n";
}
```

### Nullable

```cpp
// Nullable(String)
auto n = block[0]->As<ColumnNullableT<ColumnString>>();
for (size_t i = 0; i < n->Size(); ++i) {
    std::optional<std::string_view> v = n->At(i);
    std::cout << (v ? std::string(*v) : std::string("<null>")) << "\n";
}
```

### Tuple

```cpp
// Tuple(UInt64, String)
auto t = block[0]->As<ColumnTupleT<ColumnUInt64, ColumnString>>();
for (size_t i = 0; i < t->Size(); ++i) {
    auto [id, name] = t->At(i);                 // std::tuple<uint64_t, std::string_view>
    std::cout << id << " => " << name << "\n";
}
```

### Map

```cpp
// Map(String, UInt64)
auto m = block[0]->As<ColumnMapT<ColumnString, ColumnUInt64>>();
for (size_t i = 0; i < m->Size(); ++i) {
    auto row = m->At(i);                         // MapValueView
    for (const auto& [key, value] : row)
        std::cout << key << "=" << value << " ";
    std::cout << "\n";
}
```

### Nested types

Wrapping recurses, so nested typed columns work too:

```cpp
// Array(LowCardinality(String))
auto arr = block[0]->As<ColumnArrayT<ColumnLowCardinalityT<ColumnString>>>();
```

### `As` vs `AsStrict`

`As` returns `nullptr` on a type mismatch — always check it:

```cpp
if (auto typed = column->As<ColumnArrayT<ColumnUInt64>>()) {
    // use typed
} else {
    // wrong type
}
```

`AsStrict` throws `clickhouse::ValidationError` instead, which is handy when a mismatch is a
programming error you want to surface immediately:

```cpp
auto typed = column->AsStrict<ColumnArrayT<ColumnUInt64>>();   // throws on mismatch
```

---

## How it works (brief)

- A compile-time trait, `HasWrapMethod<T>`, detects whether `T` exposes a static `Wrap`.
- For a wrappable `T`, `As`/`AsStrict` try `dynamic_pointer_cast<T>` first, then fall back to
  `WrapColumn<T>(...)`, which calls `T::Wrap(...)`.
- `T::Wrap(...)` builds the typed wrapper by sharing the source column's internal storage
  (`shared_ptr`), recursing into nested columns as needed.
- There are non-throwing cores (`Wrap(col, ValidationError*)`) used to implement `As`, and
  throwing one-argument forms used by `AsStrict`.

---

## Limitations & gotchas

- **Check for `nullptr` (or use `AsStrict`).** `As` signals a mismatch by returning
  `nullptr`; it does not throw for a plain type mismatch.

- **Wrapping allocates.** Unless the column already is `T` (exact-match fast path), wrapping
  creates a new wrapper object (with a distinct identity from the source, though it shares
  storage). These methods are not `noexcept`: allocation can still throw `std::bad_alloc`, and
  the throwing forms raise `ValidationError` on a mismatch.

### DIFF
--- a/clickhouse/columns/column.h
+++ b/clickhouse/columns/column.h
@@ -26,26 +26,25 @@ public:
     virtual ~Column() {}
 
     /// Downcast pointer to the specific column's subtype.
+    ///
+    /// If T is a "wrappable" typed column (one exposing a static Wrap method, e.g.
+    /// ColumnArrayT/ColumnTupleT/ColumnMapT/ColumnNullableT/ColumnLowCardinalityT) and the
+    /// column is not already exactly T, this attempts to Wrap it as T (a storage-sharing
+    /// typed view). Returns nullptr when neither an exact cast nor a wrap is possible.
+    /// (Definitions are out-of-line below, after WrapColumn is declared.)
     template <typename T>
-    inline std::shared_ptr<T> As() {
-        return std::dynamic_pointer_cast<T>(shared_from_this());
-    }
+    inline std::shared_ptr<T> As();
 
-    /// Downcast pointer to the specific column's subtype.
+    /// Const overload. Unlike the non-const As(), this does NOT wrap: it only performs an
+    /// exact downcast and returns nullptr on mismatch (even for a wrappable T). Wrapping is
+    /// intentionally disabled here because it would synthesize a mutable, storage-sharing
+    /// view from a const column (requiring a const_cast), which is not const-correct.
     template <typename T>
-    inline std::shared_ptr<const T> As() const {
-        return std::dynamic_pointer_cast<const T>(shared_from_this());
-    }
+    inline std::shared_ptr<const T> As() const;
 
-    /// Downcast pointer to the specific column's subtype.
+    /// Like As(), but throws ValidationError instead of returning nullptr on failure.
     template <typename T>
-    inline std::shared_ptr<T> AsStrict() {
-        auto result = std::dynamic_pointer_cast<T>(shared_from_this());
-        if (!result) {
-            throw ValidationError("Can't cast from " + type_->GetName());
-        }
-        return result;
-    }
+    inline std::shared_ptr<T> AsStrict();
 
     /// Get type object of the column.
     inline TypeRef Type() const { return type_; }
@@ -153,6 +152,40 @@ inline std::shared_ptr<T> WrapColumn(const ColumnRef& column) {
         throw error;
     }
     return result;
+}
+
+template <typename T>
+inline std::shared_ptr<T> Column::As() {
+    if constexpr (HasWrapMethod<T>::value) {
+        if (auto exact = std::dynamic_pointer_cast<T>(shared_from_this())) {
+            return exact;
+        }
+        return WrapColumn<T>(shared_from_this(), nullptr);
+    } else {
+        return std::dynamic_pointer_cast<T>(shared_from_this());
+    }
+}
+
+template <typename T>
+inline std::shared_ptr<const T> Column::As() const {
+    // No wrapping for the const overload (see declaration): exact downcast only.
+    return std::dynamic_pointer_cast<const T>(shared_from_this());
+}
+
+template <typename T>
+inline std::shared_ptr<T> Column::AsStrict() {
+    if constexpr (HasWrapMethod<T>::value) {
+        if (auto exact = std::dynamic_pointer_cast<T>(shared_from_this())) {
+            return exact;
+        }
+        return WrapColumn<T>(shared_from_this());
+    } else {
+        auto result = std::dynamic_pointer_cast<T>(shared_from_this());
+        if (!result) {
+            throw ValidationError("Can't cast from " + type_->GetName());
+        }
+        return result;
+    }
 }
 
 }  // namespace clickhouse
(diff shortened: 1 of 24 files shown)


clickhouse/columns/column.h
```diff
@@ -104,4 +112,147 @@ class Column : public std::enable_shared_from_this<Column> {
...
+// Throwing convenience wrapper.
+template <typename T>
+inline std::shared_ptr<T> WrapColumn(const ColumnRef& column) {
+    ValidationError error;
+    auto result = WrapColumn<T>(column, &error);
+    if (!result) {
+        throw error;
+    }
+    return result;
+}
```
slabko, line 162: Moved from utils.h

joe-clickhouse (approved):

PR #552, opened 2026-08-24
Validate FixedString sizes before swapping

fallintoplace:
## What changed

- **Reject swaps between different FixedString sizes** before changing either column.
- Keep same-size swaps working as before.
- Add coverage for both valid and invalid swaps.

## Tests

- `cmake --build build --target clickhouse-cpp-ut --parallel 8`
- `./build/ut/clickhouse-cpp-ut --gtest_filter='ColumnsCase.*'`


slabko (approved):

slabko:
Thank you @fallintoplace for noticing this inconsistency!

PR #561, opened 2026-09-02
c++ complain that ciso646 is non standard, and version should be used

RoyBellingan:
Just a small 

#if __cplusplus >= 202002L

### DIFF
--- a/contrib/absl/absl/base/options.h
+++ b/contrib/absl/absl/base/options.h
@@ -70,7 +70,11 @@
 // Include a standard library header to allow configuration based on the
 // standard library in use.
 #ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


slabko:
Thank you, @RoyBellingan, for taking care of this problem.

Just as a side note, I suggest not using the vendored Abseil and disabling the use of Abseil for Int128 in general with the `-DCH_USE_ABSEIL_FOR_BIGNUM=NO` build configuration parameter.

Abseil support for Int128 will be removed in the future.

slabko (approved):

RoyBellingan:
Thank you for the kind message @slabko !

PR #562, opened 2026-09-02
Fix stale errno handling when recv returns 0

### LINKED ISSUE #487: `SocketInput::DoRead` reports stale errno on clean connection close
# `SocketInput::DoRead` reports stale errno on clean connection close (`recv() == 0`)

## Summary

When the remote peer closes the TCP connection cleanly, `recv()` returns 0. The error handling code at this point calls `getSocketErrorCode()` (which reads `errno` / `WSAGetLastError()`), but `recv()` returning 0 is **not an error** — it is an EOF indication — and the OS does **not** update `errno` in this case. The result is a `std::system_error` with a **stale, misleading error code** from a previous syscall.

## Reproduction

A minimal TCP server that accepts a connection, reads the client Hello, then immediately closes:

```cpp
int client_fd = accept(server_fd, nullptr, nullptr);
// optionally recv() the Hello
close(client_fd);
```

Client code:

```cpp
clickhouse::ClientOptions opts;
opts.SetHost("127.0.0.1");
opts.SetPort(port);
clickhouse::Client client(opts);  // throws
```

**Actual exception message:**
```
closed: Operation now in progress
```

**Expected exception message:**
```
Connection closed by peer
```

"Operation now in progress" is `EINPROGRESS` — completely unrelated to a connection close.

## Root Cause

In [`socket.cpp`, `SocketInput::DoRead`](https://github.com/ClickHouse/clickhouse-cpp/blob/master/clickhouse/base/socket.cpp):

```cpp
size_t SocketInput::DoRead(void* buf, size_t len) {
    const ssize_t ret = ::recv(s_, (char*)buf, (int)len, 0);

    if (ret > 0) {
        return (size_t)ret;
    }

    if (ret == 0) {
        throw std::system_error(getSocketErrorCode(), getErrorCategory(), "closed");
        //                      ^^^^^^^^^^^^^^^^^^^ BUG: errno is stale
    }

    throw std::system_error(getSocketErrorCode(), getErrorCategory(), "can't receive string data");
}
```

When `recv()` returns 0, the POSIX specification does not require `errno` to be set. The value of `errno` remains whatever it was from the **last syscall that failed**. In this case, the typical call sequence is:

1. `SocketConnect` → `connect()` in non-blocking mode → `errno` set to **`EINPROGRESS`**
2. `Poll()` succeeds → `getsockopt(SO_ERROR)` returns 0 → socket switched to blocking
3. `SendHello()` → `send()` succeeds → `errno` unchanged (success doesn't clear `errno`)
4. `ReceiveHello()` → `recv()` returns **0** (peer closed) → `errno` still **`EINPROGRESS`**
5. `getSocketErrorCode()` returns `EINPROGRESS` → exception says "Operation now in progress"

Depending on timing and platform, the stale value could be any previous error code, making the exception message **non-deterministic and misleading**.

## Impact

- **Debugging difficulty**: The misleading error code sends developers on the wrong path. "Operation now in progress" suggests a non-blocking socket issue, not a closed connection.
- **Error handling**: Callers catching `std::system_error` and inspecting `.code().value()` get an incorrect error code, making programmatic retry/recovery logic unreliable.
- **Non-determinism**: The stale errno value varies depending on which syscall last set it, making the bug platform- and timing-dependent.

## Suggested Fix

For the `recv() == 0` case, use a well-defined error code instead of reading `errno`:

```cpp
if (ret == 0) {
    throw std::system_error(
        ECONNRESET, getErrorCategory(), "connection closed by peer");
}
```

Or on Windows:

```cpp
if (ret == 0) {
#if defined(_win_)
    throw std::system_error(
        WSAECONNRESET, windowsErrorCategory::category(), "connection closed by peer");
#else
    throw std::system_error(
        ECONNRESET, std::system_category(), "connection closed by peer");
#endif
}
```

Alternatively, a custom error code/category could be used to distinguish a clean close (FIN) from a reset (RST), but at minimum the stale `getSocketErrorCode()` call must be removed from this path.

## Environment

- **Library**: clickhouse-cpp (all current versions)
- **Affected platforms**: All (Linux, macOS, Windows)
- **Affected file**: `clickhouse/base/socket.cpp`, `SocketInput::DoRead`

puretechteam:
Fixes #487.

When `recv()` returns 0 for a clean peer close, the socket error value is not guaranteed to be updated. The previous code could therefore surface a stale error value from an earlier system call.

Additionally handle zero length buffer and `EINTR` correctly.

### DIFF
--- a/clickhouse/base/socket.cpp
+++ b/clickhouse/base/socket.cpp
@@ -436,7 +436,11 @@ size_t SocketInput::DoRead(void* buf, size_t len) {
     }
 
     if (ret == 0) {
-        throw std::system_error(getSocketErrorCode(), getErrorCategory(), "closed");
+#if defined(_win_)
+        throw std::system_error(WSAECONNRESET, getErrorCategory(), "connection closed by peer");
+#else
+        throw std::system_error(ECONNRESET, getErrorCategory(), "connection closed by peer");
+#endif
     }
 
     throw std::system_error(getSocketErrorCode(), getErrorCategory(), "can't receive string data");
--- a/clickhouse/base/socket.cpp
+++ b/clickhouse/base/socket.cpp
@@ -1,6 +1,7 @@
 #include "socket.h"
 #include "singleton.h"
 #include "../client.h"
+#include "../exceptions.h"
 
 #include <assert.h>
 #include <stdexcept>
@@ -436,11 +437,12 @@ size_t SocketInput::DoRead(void* buf, size_t len) {
     }
 
     if (ret == 0) {
-#if defined(_win_)
-        throw std::system_error(WSAECONNRESET, getErrorCategory(), "connection closed by peer");
-#else
-        throw std::system_error(ECONNRESET, getErrorCategory(), "connection closed by peer");
-#endif
+        // Clean peer close (EOF) before the requested number of bytes
+        // arrived: the underlying `recv()` succeeded, so this is not a
+        // syscall error. The decoder expected more protocol data and the
+        // connection ended instead, so surface this as a truncated-data
+        // / protocol decoding failure rather than a system_error.
+        throw ProtocolError("connection closed by peer while reading");
     }
 
     throw std::system_error(getSocketErrorCode(), getErrorCategory(), "can't receive string data");
--- a/ut/socket_ut.cpp
+++ b/ut/socket_ut.cpp
@@ -131,6 +131,7 @@ TEST(Socketcase, connecttimeout) {
 //}
 
 #if !defined(_win_)
+#   include <cerrno>
 #   include <sys/socket.h>
 #   include <unistd.h>
 
--- a/ut/socket_ut.cpp
+++ b/ut/socket_ut.cpp
@@ -137,33 +137,21 @@ TEST(Socketcase, connecttimeout) {
 
 // Regression test for issue #487.
 //
-// On a clean peer close, `recv()` returns 0, which is EOF, not an error.
-// POSIX does NOT require `errno` to be set when `recv()` returns 0, so reading
-// `errno` at that point yields a stale value from a previous syscall. Prior
-// to the fix, `SocketInput::DoRead` surfaced that stale `errno` to the caller
-// (e.g. "Operation now in progress" if the last failing call was the
-// non-blocking `connect()`), making the exception message non-deterministic
-// and misleading.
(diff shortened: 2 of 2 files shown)


slabko:
Hi @puretechteam, could you please sign the CLA?

Also, please consider fixing your PR description.

clickhouse/base/socket.cpp
```diff
@@ -436,7 +436,11 @@ size_t SocketInput::DoRead(void* buf, size_t len) {
     }
 
     if (ret == 0) {
-        throw std::system_error(getSocketErrorCode(), getErrorCategory(), "closed");
+#if defined(_win_)
+        throw std::system_error(WSAECONNRESET, getErrorCategory(), "connection closed by peer");
+#else
+        throw std::system_error(ECONNRESET, getErrorCategory(), "connection closed by peer");
```
slabko: `ECONNRESET` = Connection reset by peer. Here, the connections are closed gracefully when unexpected. This requires a more thoughtful solution.

puretechteam: Before I revise the error value, could you confirm what error representation you would prefer here? I want to preserve the existing `std::system_error` / retry behavior rather than introduce a new error abstraction without precedent.

slabko: Honestly, I do not think it is a good idea to return `std::system_error` here because it suggests that the error occurred somewhere in a system call and that is not the case at all. The system call worked just fine and returned exactly what it was expected to return.

The error occurs at the application level during decoding because the decoder needs more data, but there is none. Maybe the server changed something, maybe a proxy dropped the connection, or maybe we are using a ClickHouse clone with a new protocol implementation.

Either way, this is a decoding error - a truncated data error, to be specific. I would use the `ProtocolError` exception type here.

puretechteam: Thanks, that makes sense. I’ll revise the `recv() == 0` path to treat the condition as a truncated protocol/decode error and use `ProtocolError` rather than `std::system_error`, while keeping actual socket/system-call failures on the existing path.

puretechteam:
I revised the `recv() == 0` handling to use `ProtocolError` for the truncated-data case, while keeping `std::system_error` for actual socket errors. I also updated the regression test to cover the new behavior. The PR is ready for re-review.

slabko (approved):

